PR #1, opened 2019-01-04
Add base package with lerna, update README

draffensperger:
This use of `lerna` parallels the [OpenCensus node](https://github.com/census-instrumentation/opencensus-node) use of it.

Having the web specific packages in a separate repo I think makes sense as it allows cleaner separation of Node vs. web functionality, and we can run separate test suites for each one. The `@opencensus/core` interfaces at least will be used in both places though so if it would be better to combine we can.


lerna.json
```diff
@@ -0,0 +1,7 @@
+{
+  "lerna": "2.11.0",
```
mayurkale22: Can we use the latest version here? https://www.npmjs.com/package/lerna

package.json
```diff
@@ -0,0 +1,41 @@
...
+    "test": "lerna run test",
+    "bootstrap": "lerna bootstrap",
+    "bump": "lerna publish",
+    "codecov": "lerna run codecov"
+  },
+  "keywords": [
+    "opencensus",
+    "browser",
+    "tracing",
+    "profiling"
```
mayurkale22: We should add ```Stats``` and ```Metrics``` here, I am gonna do the same for OC-Node repo.

lerna.json
justindsmith: We pull in `"lerna": "^3.7.1"` in the devDependencies. We should unify these at the very least, right?

package.json
draffensperger: Done.

lerna.json
draffensperger: I switched this to the latest version and also started a PR to upgrade it for `opencensus-node` as well: https://github.com/census-instrumentation/opencensus-node/pull/263

draffensperger: Realized I misunderstood the comment about unification that this involve fixing `lerna.json` here not necessarily upgrading it in `opencensus-node`. Fixed it here to match the `3.8.5` in `devDependencies`.

draffensperger:
@mayurkale22 does this look good? If so, can you approve?

PR #2, opened 2019-01-04
Initial implementation of OpenCensus agent exporter for web

draffensperger:
This takes a dev dependency on the `@opencensus/core` library so that it can implement the `Exporter` TypeScript interface provided by it. It uses webpack to shake out the current Node-specific dependencies of `@opencensus/core`.

This exporter sends data via HTTP/JSON using the browser's XMLHttpRequest API. This relies on the `grpc-gateway` functionality added to the OpenCensus agent in https://github.com/census-instrumentation/opencensus-service/pull/270.

This uses Jasmine for tests, which is nice for browser-based code since it provides mocks and expectations out of the box (can't use the Node's `assert.ok` in browser Karma tests). So having the tests pass depends on https://github.com/census-instrumentation/opencensus-node/pull/254 to make the Mocha types not be included in the `@opencensus/core` release code.

### DIFF
--- a/packages/opencensus-web-exporter-ocagent/src/adapters.ts
+++ b/packages/opencensus-web-exporter-ocagent/src/adapters.ts
@@ -1,0 +1,186 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as coreTypes from '@opencensus/core';
+import * as apiTypes from './api-types';
+
+/**
+ * Converts a RootSpan type from @opencensus/core to the Span JSON structure
+ * expected by the OpenCensus Agent's HTTP/JSON (grpc-gateway) API.
+ */
+export function adaptRootSpan(rootSpan: coreTypes.RootSpan): apiTypes.Span[] {
+  const adaptedSpans: apiTypes.Span[] = rootSpan.spans.map(adaptSpan);
+  adaptedSpans.unshift(adaptSpan(rootSpan));
+  return adaptedSpans;
+}
+
+function adaptString(value: string): apiTypes.TruncatableString {
+  return {value};
+}
+
+/** Converts hexadecimal string to base64 string. */
+function hexToBase64(hexStr: string): string {
+  const hexStrLen = hexStr.length;
+  let hexAsciiCharsStr = '';
+  for (let i = 0; i < hexStrLen; i += 2) {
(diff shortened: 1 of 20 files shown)


packages/opencensus-web-exporter-ocagent/src/api-types.ts
```diff
@@ -0,0 +1,531 @@
...
+ * can also contain multiple root spans, or none at all. Spans do not need to be
+ * contiguous - there may be gaps or overlaps between spans in a trace.  The
+ * next id is 16. TODO(bdrutu): Add an example.
+ */
+export interface Span {
+  /**
+   * A unique identifier for a trace. All spans from the same trace share the
+   * same `trace_id`. The ID is a 16-byte array.  This field is required.
+   */
+  traceId?: string;
```
justindsmith: Since the field is required, should this property (and the `spanId` below) be required?

packages/opencensus-web-exporter-ocagent/test/test-exporter-buffer.ts
```diff
@@ -0,0 +1,72 @@
...
+    buffer.addToBuffer(SPAN2);
+    expect(exporter.publish).toHaveBeenCalledWith([SPAN1, SPAN2]);
+
+    // Adding a third span should not trigger another flush since the buffer was
+    // cleared after the second spand.
+    buffer.addToBuffer(SPAN3);
+    expect(exporter.publish).toHaveBeenCalledTimes(1);
+  });
+
+  it('flushes root spans when it reaches a buffer size maximum', () => {
```
justindsmith: "buffer size" should be "buffer timeout"

draffensperger: Done.

packages/opencensus-web-exporter-ocagent/webpack.config.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright 2018, OpenCensus Authors
```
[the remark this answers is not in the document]
draffensperger: Done.

packages/opencensus-web-exporter-ocagent/src/adapters.ts
```diff
@@ -0,0 +1,186 @@
...
+  }
+}
+
+function adaptValue(value: boolean|string|number): apiTypes.AttributeValue {
+  const valType = typeof value;
+  if (valType === 'boolean') {
+    return {boolValue: value as boolean};
+  }
+  if (valType === 'number' && value === Math.floor(value as number)) {
+    return {intValue: String(value)};
```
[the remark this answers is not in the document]
draffensperger: Good point - the `intValue` would be written as a string from a read API that was based on int64 protobuf values because the JS `number` can't accurately hold all 64-bit ints.

But here we are going the other way so we are fine to keep it as a number and I tested with the OpenCensus agent grpc-gateway and it works fine as a number.

I noticed now that there is support for a `doubleValue` attribute type since https://github.com/census-instrumentation/opencensus-proto/pull/98/files . That seems like a more exact match with the JS `number` type so I changed it to that here and in the tests. Maybe if this eventually supports big ints it could use `intValue`, but full browser support for that is still a while off.

draffensperger:
@mayurkale22 I ran into some git troubles (have not used GitHub in a while), but the net result is that this now has a clean commit and also includes the top level `package.json` file that got lost in my previous PR. I realize this is bad as it loses the history and will try to be cleaner about this next time.

Could you take a look again when you get a chance?

PR #3, opened 2019-01-17
Fixes to prepare for enabling CircleCI

draffensperger:
This fixes a couple of issues that should allow the CircleCI build to pass once we enable it.


draffensperger:
Well, `opencensus-node` uses CircleCI, so I figured we should follow that. Travis is fine too. Which would you tend to recommend?

mayurkale22:
> Well, opencensus-node uses CircleCI, so I figured we should follow that. Travis is fine too. Which would you tend to recommend?

Ok, SGTM. Make sure we use CircleCI 2.0 (https://circleci.com/blog/sunsetting-1-0/).

draffensperger:
OK, sure. The config file has `version: 2.0` following the opencensus-node config.

PR #8, opened 2019-02-08
Add url parsing utility

draffensperger:
This will be used by the upcoming browser timing instrumentation package to specify the various HTTP attributes from the [OpenCensus HTTP specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#attributes) such as host and path.

### DIFF
--- a/packages/opencensus-web-core/src/common/url-util.ts
+++ b/packages/opencensus-web-core/src/common/url-util.ts
@@ -1,7 +1,7 @@
 /**
  * Copyright 2019, OpenCensus Authors
  *
- * Licensed under the Apache License, Version 2.0 the "License";
+ * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
--- a/packages/opencensus-web-core/src/common/url-util.ts
+++ b/packages/opencensus-web-core/src/common/url-util.ts
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Interface to represent the components of a parsed URL.
+ * The field comments below reference this sample URL:
+ *    https://example.com:3000/pathname/?search=test#hash
+ */
+export interface ParsedUrl {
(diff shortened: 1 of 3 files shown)


packages/opencensus-web-core/test/test-url-util.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('parseUrl', () => {
+  it('parses a URL into its various components', () => {
+    const parsedUrl =
+        parseUrl('https://example.com:3000/pathname/?search=test#hash');
+
+    expect(parsedUrl.protocol).toBe('https');
+    expect(parsedUrl.hostname).toBe('example.com');
+    expect(parsedUrl.port).toBe('3000');
+    expect(parsedUrl.host).toBe('example.com:3000');
+    expect(parsedUrl.origin).toBe('http://example.com:3000');
```
mayurkale22: This should be ```https://example.com:3000```, right?

packages/opencensus-web-core/test/test-url-util.ts
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import 'jasmine';
+import {parseUrl} from '../src/common/url-util';
+
+describe('parseUrl', () => {
+  it('parses a URL into its various components', () => {
+    const parsedUrl =
```
mayurkale22: Can we add one more test without port? something like ```http://example.com/path1/path2?search=test&format=json```

packages/opencensus-web-core/test/test-url-util.ts
```diff
@@ -0,0 +1,34 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
mayurkale22: ```(the "License")``` vs ```the "License"```, we are using both.

packages/opencensus-web-core/src/common/url-util.ts
```diff
@@ -0,0 +1,46 @@
...
+ * The field comments below reference this sample URL:
+ *    https://example.com:3000/pathname/?search=test#hash
+ */
+export interface ParsedUrl {
+  /** Protocol of the URL. E.g. https */
+  protocol: string;
+  /** Host name only (no port). E.g. example.com */
+  hostname: string;
+  /** Port number if explicitly specified (else empty string). E.g. 3000 */
+  port: string;
```
mayurkale22: ```number``` instead of ```string```?

packages/opencensus-web-core/test/test-url-util.ts
draffensperger: Sure, done.

packages/opencensus-web-core/src/common/url-util.ts
draffensperger: Well, because this is in a browser context where I want to keep code size small, I'm implementing this by generating an `<a>` tag, which uses `string` for the port (and leaves it as an empty string if it is not explicitly specified). I added a comment to explain that.

packages/opencensus-web-core/test/test-url-util.ts
draffensperger: Fixed here to use parens, thanks!

PR #11, opened 2019-02-08
Export browser performance timestamps for time events

draffensperger:
This makes the OpenCensus web agent exporter support browser performance times in the message event and annotation `timestamp` field. The field is a number and for OpenCensus Node represents an epoch time.

However, JS doubles in epoch time have less accuracy than small JS numbers, and it's more convenient for the instrumentation to be able to set performance times as the `timestamp` field since that's what the browser performance API and `performance.now()` return.

This handles both cases by checking the value of the numberic field to determine whether it's an epoch time or performance time - maybe a little hacky but it works and is compatible with both uses.


draffensperger:
@mayurkale22 could you take a look at this when you get a chance?

PR #12, opened 2019-02-12
Performance timing recording and span conversion

draffensperger:
This adds utility functions to begin recording browser performance timings and to convert them into a `RootSpan` for the overall time between the initial navigation and when the `load` event fires including resource and long task timings along the way.

Nothing is exported in `index.ts` yet because I want to get this integrated into a working example to figure out what is the most useful interface.

### DIFF
--- a/packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
+++ b/packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
@@ -1,0 +1,126 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Annotation, ATTRIBUTE_HTTP_URL, ATTRIBUTE_HTTP_USER_AGENT, ATTRIBUTE_LONG_TASK_ATTRIBUTION, ATTRIBUTE_NAV_TYPE, parseUrl, RootSpan, Span, Tracer} from '@opencensus/web-core';
+import {GroupedPerfEntries} from './perf-recorder';
+import {PerformanceLongTaskTiming, PerformanceNavigationTimingExtended} from './perf-types';
+import {getResourceSpan} from './resource-span';
+import {annotationsForPerfTimeFields} from './util';
+
+/**
+ * These are properties of PerformanceNavigationTiming that will be turned
+ * into span annotations on the navigation span.
+ */
+const NAVIGATION_TIMING_EVENTS = [
+  'domLoading',
+  'domInteractive',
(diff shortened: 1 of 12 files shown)


packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
```diff
@@ -0,0 +1,126 @@
...
+export function getInitialLoadRootSpan(
+    tracer: Tracer, perfEntries: GroupedPerfEntries,
+    navigationFetchSpanId: string, traceId: string): RootSpan {
+  const navTiming = perfEntries.navigationTiming;
+  const navigationUrl = navTiming ? navTiming.name : location.href;
+  const parsedNavigationUrl = parseUrl(navigationUrl);
+  const navigationPath = parsedNavigationUrl.pathname;
+  const root = new RootSpan(
+      tracer,
+      {name: `Nav.${navigationPath}`, spanContext: {spanId: '', traceId}});
```
mayurkale22: Should we specify ```SpanKind``` here? and a comment for empty ```spanId```.

packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
```diff
@@ -0,0 +1,126 @@
...
+        getNavigationFetchSpan(navTiming, navigationUrl, navigationFetchSpanId);
+    root.spans.push(navFetchSpan);
+  }
+
+  const resourceSpans = perfEntries.resourceTimings.map(getResourceSpan);
+  const longTaskSpans = perfEntries.longTasks.map(getLongTaskSpan);
+
+  root.spans = root.spans.concat(resourceSpans, longTaskSpans);
+  for (const span of root.spans) {
+    span.traceId = traceId;
```
mayurkale22: Is it possible to pass ```traceId``` and ```root.id``` while constructing spans (```resourceSpans``` and ```longTaskSpans```), instead of looping here again?

packages/opencensus-web-instrumentation-perf/test/test-initial-load-root-span.ts
```diff
@@ -0,0 +1,304 @@
...
+    expect(root.id).toMatch(SPAN_ID_REGEX);
+    expect(root.traceId).toBe(traceId);
+    expect(root.startPerfTime).toBe(0);
+    expect(root.endPerfTime).toBe(20985.30000000028);
+    expect(root.attributes).toEqual(EXPECTED_ROOT_ATTRIBUTES);
+    expect(root.annotations).toEqual(EXPECTED_ROOT_ANNOTATIONS);
+
+    expect(root.spans.length).toBe(3);
+
+    const navigationFetchSpan = root.spans[0];
```
mayurkale22: For unpacking multiple spans, can we use array destructuring here?

packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
draffensperger: Sure, added an unspecified span kind here and a comment about the empty span ID (it will become the parent span ID).

packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
draffensperger: Sure, I added those parameters to the relevant functions and removed this loop.

packages/opencensus-web-instrumentation-perf/test/test-initial-load-root-span.ts
draffensperger: Done, good thought.

PR #15, opened 2019-02-28
Add empty opencensus-web-all package for generating JS bundles

draffensperger:
This will pull together the main packages of OpenCensus Web and enable creation of various JS bundle distributions to give the option of including it web applications via `<script>` tags rather than as a library as part of the application build.

This is similar to the [opencensus-instrumentation-all](https://github.com/census-instrumentation/opencensus-node/tree/master/packages/opencensus-instrumentation-all) package in that it will pull in the primary functionality of other packages in the library.

### DIFF
--- a/packages/opencensus-web-all/tsconfig.json
+++ b/packages/opencensus-web-all/tsconfig.json
@@ -1,0 +1,21 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "pretty": true,
+    "module": "commonjs",
+    "target": "es5",
+    "lib": ["es5", "dom", "es6"],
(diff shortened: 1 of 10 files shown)


packages/opencensus-web-all/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "pretty": true,
+    "module": "commonjs",
+    "target": "es5",
+    "lib": ["es5", "dom", "es6"],
+    "strictNullChecks": true,
```
mayurkale22: We can add ```noUnusedLocals``` ?

draffensperger: Sure, done.

PR #23, opened 2019-03-13
Remove @opencensus/core dep for ocagent exporter

draffensperger:
I'm planning to make this library buildable with Bazel (see [bazelbuild/rules_typescript](https://github.com/bazelbuild/rules_typescript)) and the Closure compiler.

As a prerequisite for that, I'm planning to refactor the way that OpenCensus Web ensures that it implements the `@opencensus/core` interfaces so that the Bazel build can be cleaner and the Closure optimization can rename a wider group of property names (including those in `@opencensus/core`). My plan is as follows:

- Make it so that only `@opencensus/web-core` depends on `@opencensus/core` (this PR)
- Make `@opencensus/core` have its own copies of the trace type interfaces, but include unit tests to ensure that its `Span` class also correctly implements the `Span` interface in `@opencensus/core` (this can be done by making a derived class that implements the interface)
- Then add Bazel + Closure build rules


packages/opencensus-web-exporter-ocagent/src/adapters.ts
```diff
@@ -29,7 +28,7 @@ const RECENT_EPOCH_MS = 1500000000000;  // July 13, 2017.
  * Converts a RootSpan type from @opencensus/core to the Span JSON structure
```
mayurkale22: Should we change this comment from ```@opencensus/core``` -> ```@opencensus/web-core```?

draffensperger: Good catch, done.

draffensperger:
The Closure compiler has some more advanced optimization/minification techniques that should result in smaller bundle sizes for use in `<script>` tags. The Bazel build system allows for incremental builds and is also more similar to Google's internal build system and so would make it easier to use this internally for Google projects as well. 

See [Angular BuildTools Convergence](https://github.com/angular/angular/issues/19058) for more on Bazel for TypeScript.

draffensperger:
The reason I wanted to do this sooner rather than later is that the Closure compiler also restricts certain more dynamic uses of JS due to how it renames properties, so wanted to set up some build rules in CI to make sure future changes would be Closure compiler safe.

PR #26, opened 2019-03-21
Script to copy types from `@opencensus/core`

draffensperger:
This script copies in the type files from `@opencensus/core` to avoid the dependency issues described in https://github.com/census-instrumentation/opencensus-web/blob/master/packages/opencensus-web-types/README.md

This also brings in the polyfill for the `NodeJS.EventEmitter` type so that no dependency on `@types/node` is needed.

PR #27, opened 2019-03-21
Pull in @opencensus/core types using `copytypes` npm script

draffensperger:
This is the result of running `npm run copytypes` using the script written in https://github.com/census-instrumentation/opencensus-web/pull/26

The code itself is from `@opencensus/core` at the `v0.0.9` git tag.

### DIFF
--- a/packages/opencensus-web-types/scripts/copy-types.js
+++ b/packages/opencensus-web-types/scripts/copy-types.js
@@ -1,0 +1,140 @@
+/**
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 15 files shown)


mayurkale22:
> This is the result of running `npm run copytypes` using the script written in #26 

Just curious, Do we need to run this script manually everytime when there is new version of core?

draffensperger:
Yes, we will need to modify the `package.json` file with the new git tag for the release and run `npm run copyfiles`. See https://github.com/census-instrumentation/opencensus-web/blob/7271b946bfb87043ac73003944716092615d9ec8/packages/opencensus-web-types/package.json#L11

That process is explained in the [package README](https://github.com/census-instrumentation/opencensus-web/blob/master/packages/opencensus-web-types/README.md).

Because this is based on the git code rather than what's in NPM, we could also potentially sync types more often, which would help the syncs to be smaller. When the OpenCensus Web library is released though, it should have its types align with the latest `@opencensus/core` release so that there is actually a coherence with a build in NPM.

PR #29, opened 2019-03-22
Remove clock sync and initial load trace ID assignment code

draffensperger:
Per conversation with @bogdandrutu this will simplify the way that server side code interacts with OpenCensus web, since it will only need to specify an agent endpoint and programmatically generated `traceparent` header value.

The actual use of the `traceparent` header will be done in a follow up PR since it will depend on a `tracecontext` propagation implementation for OC web.


draffensperger:
Hi @mayurkale22 could you take a look at this when you get a chance?

PR #31, opened 2019-03-23
Implement tracecontext propagation format

draffensperger:
This is quite similar to the Node package `@opencensus/propagation-tracecontext`, but this does not make use of Node-specific libraries (for trace ID generation). It can makes a couple simplifications and having this as a separate package for web could enable additional optimizations once we use the Closure compiler, e.g. not doing validation in the minified production builds.

### DIFF
--- a/packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
+++ b/packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
@@ -64,7 +64,7 @@ export function spanContextToTraceParent(spanContext: SpanContext): string {
  * Propagates span context through Trace Context format propagation.
  *
  * Based on the Trace Context specification:
- * https://w3c.github.io/distributed-tracing/report-trace-context.html
+ * https://www.w3.org/TR/trace-context/
  */
 export class TraceContextFormat implements Propagation {
   /**
@@ -111,7 +111,6 @@ export class TraceContextFormat implements Propagation {
       traceId: randomTraceId(),
       spanId: randomSpanId(),
       options: DEFAULT_OPTIONS,
-      traceState: undefined,
     };
   }
 }
--- a/packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
+++ b/packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
@@ -69,8 +69,8 @@ export function spanContextToTraceParent(spanContext: SpanContext): string {
 export class TraceContextFormat implements Propagation {
   /**
    * Gets the trace context from a request headers. If there is no trace context
-   * in the headers, or if the parsed `traceId` or `spanId` is invalid, an empty
-   * context is returned.
+   * in the headers, or if the parsed `traceId` or `spanId` is invalid, null
+   * returned.
    */
   extract(getter: HeaderGetter): SpanContext|null {
     const traceParentHeader = getter.getHeader(TRACE_PARENT);
--- a/packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
+++ b/packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
@@ -1,0 +1,117 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -0,0 +1,117 @@
...
+export function spanContextToTraceParent(spanContext: SpanContext): string {
+  return `00-${spanContext.traceId}-${spanContext.spanId}-0${
+      (spanContext.options || DEFAULT_OPTIONS).toString(16)}`;
+}
+
+/**
+ * Propagates span context through Trace Context format propagation.
+ *
+ * Based on the Trace Context specification:
+ * https://w3c.github.io/distributed-tracing/report-trace-context.html
```
mayurkale22: Looks like this link is broken..

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -0,0 +1,117 @@
...
+ */
+export class TraceContextFormat implements Propagation {
+  /**
+   * Gets the trace context from a request headers. If there is no trace context
+   * in the headers, or if the parsed `traceId` or `spanId` is invalid, an empty
+   * context is returned.
```
mayurkale22: `s/an empty context/null`?

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Parses a traceparent header value into a SpanContext object, or null if the
+ * traceparent value is invalid.
+ */
+export function traceParentToSpanContext(traceParent: string): SpanContext|
+    null {
+  const match = traceParent.match(TRACE_PARENT_REGEX);
+  if (!match) return null;
+  const parts = traceParent.split('-');
+  const traceId = parts[1];
```
mayurkale22: Should we validate the version here? `parts[0]`?

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Parses a traceparent header value into a SpanContext object, or null if the
+ * traceparent value is invalid.
+ */
+export function traceParentToSpanContext(traceParent: string): SpanContext|
+    null {
+  const match = traceParent.match(TRACE_PARENT_REGEX);
+  if (!match) return null;
+  const parts = traceParent.split('-');
```
mayurkale22: Does it make sense to validate the length of parts?

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -0,0 +1,117 @@
...
+   * Generate SpanContext.
+   *
+   * Context parts are based on section 2.2.2 of TraceContext spec.
+   */
+  generate(): SpanContext {
+    return {
+      traceId: randomTraceId(),
+      spanId: randomSpanId(),
+      options: DEFAULT_OPTIONS,
+      traceState: undefined,
```
mayurkale22: This(`traceState`) is an optional param, I think it is safe to remove instead of assigning `undefined`. I can make similar change in OC Node repo.

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
draffensperger: Done

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
draffensperger: Thanks! Replaced it with https://www.w3.org/TR/trace-context/

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
draffensperger: Well, the version is validated indirectly via the regex above. It seems like based on the Node library (https://github.com/census-instrumentation/opencensus-node/blob/ac0ef77b8c1a3c0c33642ea575aacd7e965dcd27/packages/opencensus-propagation-tracecontext/src/validators.ts#L58) that we don't require the version to match a specific value but rather attempt to parse the header whatever the version is.

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
draffensperger: Yes, though the lengths are validated implicitly by matching the `TRACE_PARENT_REGEX`

draffensperger:
I think I addressed all the comments. @mayurkale22 is this ready to merge?

packages/opencensus-web-propagation-tracecontext/src/tracecontext-format.ts
mayurkale22: Ok.

PR #34, opened 2019-03-26
Use traceparent from window for initial load context

draffensperger:
This modifies the initial load span export logic to look for a `traceparent` property on `window` to set the trace and span ID for the initial load. If that is missing, a new random span context will be generated - for now that new context is always marked as sampled though a sampling rate configuration could be added later.

### DIFF
--- a/packages/opencensus-web-all/src/initial-load-context.ts
+++ b/packages/opencensus-web-all/src/initial-load-context.ts
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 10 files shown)


packages/opencensus-web-all/src/initial-load-context.ts
```diff
@@ -0,0 +1,47 @@
...
+    return randomSampledSpanContext();
+  }
+  return spanContext;
+}
+
+function randomSampledSpanContext() {
+  return {
+    traceId: randomTraceId(),
+    spanId: randomSpanId(),
+    options: 0x1,  // Sampled
```
mayurkale22: Are we planning to change this later? If yes, can you please add TODO here. Otherwise we should change function name to something like `alwaysSampledSpanContext`.

packages/opencensus-web-propagation-tracecontext/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@opencensus/web-propagation-tracecontxt",
+  "name": "@opencensus/web-propagation-tracecontext",
```
mayurkale22: 👍

packages/opencensus-web-all/src/initial-load-context.ts
draffensperger: I may change this at some point in the future, but I renamed it to `alwaysSampledSpanContext` to be clearer. (I'm not sure yet whether sampling should be handled by the client or the agent if the server doesn't write a `window.traceparent` value in the HTML response and I may want to focus on other areas of the library first).

Issue #49, opened 2019-04-15
Write library language field for OC web to OC agent


draffensperger:
This depends on https://github.com/census-instrumentation/opencensus-proto/pull/198

draffensperger:
In order for this to be useful to end users, it also requires https://github.com/census-instrumentation/opencensus-service/issues/525 for the OC agent to write the library value to a span label.

PR #57, opened 2019-04-20
Sent OC Web library type and version to OC agent

draffensperger:
This will enable the OpenCensus Agent's Stackdriver Exporter to write the `g.co/agent` label to the exported traces.

Fixes https://github.com/census-instrumentation/opencensus-web/issues/49

### DIFF
--- a/packages/opencensus-web-exporter-ocagent/test/test-ocagent.ts
+++ b/packages/opencensus-web-exporter-ocagent/test/test-ocagent.ts
@@ -18,6 +18,8 @@ import * as webCore from '@opencensus/web-core';
 
 import {OCAgentExporter} from '../src';
 import * as apiTypes from '../src/api-types';
+import {EXPORTER_VERSION} from '../src/version';
+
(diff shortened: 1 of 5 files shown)


packages/opencensus-web-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -107,7 +107,11 @@ describe('OCAgentExporter', () => {
       node: {
         identifier: {hostName: window.location.host},
         serviceInfo: {name: 'testService'},
-        libraryInfo: {exporterVersion: '0.0.1'},
+        libraryInfo: {
+          language: 10,
+          coreLibraryVersion: '0.0.1',
```
songy23: Consider using `VERSION` and `EXPORTER_VERSION` instead of hard code. Otherwise this test needs to be updated on every new release.

draffensperger: Good point! Updated to use the constants for the versions.

Issue #46, opened 2019-04-15
Provide option to set a sampling rate if no sampling decision given from server


draffensperger:
Currently if there is no `traceparent` global variable set by a `<script>` tag sent by the server, then the library will trace every request.

In order to make this server-side configuration optional, the library should support a lightweight way to set a sampling rate by setting an `ocSampleRate` global variable that would configure the sampling rate in the absence of a `traceparent` header.

PR #58, opened 2019-04-20
Add sample rate option and update README for it

draffensperger:
This will enable the use case when someone wants to only sample some requests but doesn't make their backend server send back the `traceparent` header.

Fixes https://github.com/census-instrumentation/opencensus-web/issues/46

### DIFF
--- a/packages/opencensus-web-all/src/initial-load-context.ts
+++ b/packages/opencensus-web-all/src/initial-load-context.ts
@@ -25,7 +25,7 @@ const windowWithOcwGlobals = window as WindowWithOcwGlobals;
  * The default trace sampling rate if no `traceparent` and no `ocSampleRate`
  * are specified on the `window`.
  */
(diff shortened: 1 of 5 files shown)


packages/opencensus-web-all/src/initial-load-context.ts
```diff
@@ -21,27 +21,37 @@ import {WindowWithOcwGlobals} from './types';
 
 const windowWithOcwGlobals = window as WindowWithOcwGlobals;
 
+/**
+ * The default trace sampling rate if no `traceparent` and no `ocSampleRate`
+ * are specified on the `window`.
+ */
+const DEFAULT_SAMPLE_RATE = 0.001;
```
songy23: FYI in other languages the default is 1/10000.

draffensperger: Thanks for pointing that out. Updated to `0.0001` (1/10000) to match the other languages.

PR #75, opened 2019-05-13
Update @opencensus/web-types  with @opencensus/core

crdgonzalezca:
The last commit this syncs to is [36f63e6](https://github.com/census-instrumentation/opencensus-node/commit/36f63e6dee9e6ad9af5ece9c5458b47786d41f06#diff-3f8af03ea7f8295217dd1ec432ceda36)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Add support for object(```SpanOptions```) as an argument for ```startChildSpan``` function, similar to ```startRootSpan```.
```
draffensperger: For single line code snippets you can just use a single backtick. You only need the triple backtick when you want multi-line code snippets

packages/opencensus-web-types/src/exporters/types.ts
```diff
@@ -56,6 +56,9 @@ export interface StatsEventListener {
    * batched data to backend.
    */
   start(): void;
+
+  /** Stops the exporter. */
+  stop(): void;
```
draffensperger: Did you need to implement this somewhere?

crdgonzalezca: As we are not using Stats, we are not actually implementing this interface.

PR #80, opened 2019-05-17
Sync @opencensus/web-types to @opencensus/core v0.0.12

crdgonzalezca:


packages/opencensus-web-core/src/trace/model/tracer-base.ts
```diff
@@ -52,6 +52,9 @@ export class TracerBase implements webTypes.TracerBase {
    */
   readonly activeTraceParams = {};
 
+  /** A configuration for starting the tracer */
+  private config!: webTypes.TracerConfig;
```
draffensperger: Is there a default value we can give to this? Maybe just an empty object so `rivate config: webTypes.TracerConfig = {}`?

packages/opencensus-web-core/src/trace/model/tracer-base.ts
```diff
@@ -76,6 +80,13 @@ export class TracerBase implements webTypes.TracerBase {
    */
   startRootSpan<T>(options: webTypes.TraceOptions, fn: (root: Span) => T): T {
     const rootSpan = new RootSpan(this, options);
+    // Add default attributes
+    const defaultAttributes = this.config && this.config.defaultAttributes;
+    if (defaultAttributes) {
+      Object.keys(defaultAttributes).forEach(key => {
```
draffensperger: Could you switch this to a `for (const of` loop? That is better for the Closure compiler, which I'm hoping we will eventually use.

packages/opencensus-web-core/src/trace/model/tracer-base.ts
```diff
@@ -108,10 +119,18 @@ export class TracerBase implements webTypes.TracerBase {
...
+    let rootSpan = new Span();
     if (options && options.childOf) {
-      span = options.childOf as Span;
+      rootSpan = options.childOf as Span;
+    }
+    const span = rootSpan.startChildSpan(options);
+    // Add default attributes
+    const defaultAttributes = this.config && this.config.defaultAttributes;
+    if (defaultAttributes) {
+      Object.keys(defaultAttributes).forEach(key => {
```
draffensperger: Same here on the `for (const .. of` loop

PR #85, opened 2019-05-24
Setup InteractionTracker to log when a click occurs.

crdgonzalezca:
*Setup @opencensus/web-instrumentation-zone with InteractionTracker class to start the monkey-patch on Zone.js.

*Add dependencies in @opencensus/web-instrumentation-zone and user_interaction example.

*The InteractionTracker logs when the button is clicked, this is when a task is running, also, logs when a task is scheduled and canceled.

*Setup bundle configuration for @opencensus/web-instrumentation-zone

### DIFF
--- a/examples/user_interaction/client/src/index.js
+++ b/examples/user_interaction/client/src/index.js
@@ -18,6 +18,10 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 import { exportRootSpanAfterLoadEvent } from '@opencensus/web-all';
+import { startInteractionTracker } from '@opencensus/web-instrumentation-zone';
+
+// Necessary import for @opencensus/web-instrumentation-zone
+import { Zone, ZoneType, Task } from 'zone.js';
 
 ReactDOM.render(<App />, document.getElementById('root'));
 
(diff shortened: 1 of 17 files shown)


examples/user_interaction/client/src/index.js
```diff
@@ -18,6 +18,10 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 import { exportRootSpanAfterLoadEvent } from '@opencensus/web-all';
+import { startInteractionTracker } from '@opencensus/web-instrumentation-zone';
+
+// Necessary import for @opencensus/web-instrumentation-zone
+import { Zone, ZoneType, Task } from 'zone.js';
```
draffensperger: I'm a bit surprised this line is needed here. What happens if you remove it?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -0,0 +1,69 @@
...
+  target: HTMLElement;
+};
+
+export class InteractionTracker {
+  constructor() {
+    const runTask = Zone.prototype.runTask;
+    Zone.prototype.runTask = function(
+      task: AsyncTask,
+      // tslint:disable-next-line:no-any
+      applyThis: any,
```
draffensperger: Could you change this to `unknown` here and below and then remove the tslint error?

examples/user_interaction/client/src/index.js
crdgonzalezca: I already did that but compilation fails.

draffensperger: OK, we can leave it in there for now. Longer term we will want to think about how to bring in Zone.js cleanly, but for now if this is needed we can keep it.

PR #87, opened 2019-05-31
Create a new zone when click is detected

crdgonzalezca:
This creates a new zone when a click is detected and runs the coming tasks on that new zone.

Also, modify the user_interaction example to do the requests using XMLHttpRequest.


examples/user_interaction/client/src/App.js
```diff
@@ -23,47 +23,49 @@ class App extends React.Component {
     this.state = { pi: { time: 0, value: "unknown" }, prime_numbers: { time: 0, value: [] } };
     this.handleClick = this.handleClick.bind(this);
     this.runSecondTask = this.runSecondTask.bind(this);
-    this.http = require('http');
```
draffensperger: Now that you aren't using `http` here, can you remove the `axios` dependency in the `package.json` file?

examples/user_interaction/client/src/App.js
```diff
@@ -23,47 +23,49 @@ class App extends React.Component {
...
-      let data = [];
-      response.on('data', chunk => data.push(chunk));
-      response.on('end', this.runSecondTask);
-    });
+    console.log("Entering handle click.");
+
+    this.runFirstTask();
+  }
+
+  runFirstTask() {
```
draffensperger: What would you think about giving this a more desriptive name like `callSleepApi` or similar?

examples/user_interaction/client/src/App.js
```diff
@@ -23,47 +23,49 @@ class App extends React.Component {
...
+    xhr.onreadystatechange = () => {
+      if (xhr.readyState === XMLHttpRequest.DONE) {
+        this.runSecondTask();
+      }
+    };
+    xhr.open('GET', this.host + "/sleep");
+    xhr.send();
   }
 
   runSecondTask() {
```
draffensperger: Similarly, what would you think about calling this `callPrimeNumbersApi` or similar?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -29,39 +31,93 @@ export type AsyncTask = Task & {
...
+      let taskZone = this;
+      if (isTrackedElement(task)) {
         console.log('Click detected');
+
+        const zoneName = randomSpanId();
+        const tracingZone = Zone.root.fork({
+          name: zoneName,
+          properties: {
+            isTracingZone: true,
+            tracingId: randomTraceId(),
```
draffensperger: Nit: we use `traceId` rather than `tracingId`

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -29,39 +31,93 @@ export type AsyncTask = Task & {
...
+      const time = Date.now();
+
+      console.warn('Running task');
+      console.log(task.zone);
+
+      let taskZone = this;
+      if (isTrackedElement(task)) {
         console.log('Click detected');
+
+        const zoneName = randomSpanId();
```
draffensperger: What would you think about using the `traceId` as the zoneName rather than the spanId? Generally spans are within a given trace, so the traceId would be the top level ID. Also, since it would just be the trace Id, could we call the zone key simply `traceId`?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -29,39 +31,93 @@ export type AsyncTask = Task & {
 };
 
 export class InteractionTracker {
+  private readonly tracingZones: { [index: string]: Zone } = {};
```
draffensperger: What's the purpose of storing these Zones indexed by unique ID?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -29,39 +31,93 @@ export type AsyncTask = Task & {
...
+            isTracingZone: true,
+            tracingId: randomTraceId(),
+          },
+        });
+
+        interactionTracker.tracingZones[zoneName] = tracingZone;
+
+        // Change the zone task.
+        // tslint:disable:no-any
+        (task as any)._zone = tracingZone;
```
draffensperger: Rather than using `any` here, could you create a new interface something like this:
```
/** Interface that allows access to the private `_zone` property of a Zone.js Task. */
interface TaskWithPrivateZone {
  _zone: Zone;
}
```

Then you can do `(task as TaskWithPrivateZone)._zone = tracingZone`, but with a bit higher type safety and without needing the `tslint:disable:no-any`

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -29,39 +31,93 @@ export type AsyncTask = Task & {
...
+        taskZone = tracingZone;
+        console.log('New zone:');
+        console.log(taskZone);
+      } else {
+        // If we already are in a tracing zone, just run the task in our tracing zone.
+        if (task.zone && interactionTracker.tracingZones[task.zone.name]) {
```
draffensperger: Could you just check the `isTracingZone` property on the Zone data, which would then make it so you don't need to store the object of `tracingZones`? Or is that not easily accessible?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -29,39 +31,93 @@ export type AsyncTask = Task & {
...
-        return cancelTask.call(this as {}, task);
+        return cancelTask.call(taskZone as {}, task);
       } finally {
       }
     };
   }
 }
+
+function isTrackedElement(task: AsyncTask): boolean {
+  const eventType = task.eventName;
```
draffensperger: Optional: you could rewrite this function as a single line `return task.eventName && task.eventName === 'click';`

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: Instead of creating a new Interface, I added this property in AsyncTask type.

examples/user_interaction/client/src/App.js
crdgonzalezca: They were not present in the dependencies. However, the server had axios and I removed it.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -26,42 +28,95 @@ export type AsyncTask = Task & {
   data: AsyncTaskData;
   eventName: string;
   target: HTMLElement;
+  // Allows access to the private `_zone` property of a Zone.js Task.
+  _zone: Zone;
 };
 
 export class InteractionTracker {
+  // Store the Zones indexed by unique ID in order to check current zones.
+  // private readonly tracingZones: { [index: string]: Zone } = {};
```
draffensperger: Could you remove this and the other commented out code below if it's no longer needed?

PR #88, opened 2019-06-03
Intercept multiple events triggered by the same DOM element and create only one Zone.

crdgonzalezca:
When a button is clicked it may trigger click event from the _button_ and the _document_ but in the end both events correspond to the same user interaction and they must be tracked in the same Zone. This PR fixes this.

### DIFF
--- a/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
@@ -33,7 +33,12 @@ export type AsyncTask = Task & {
 };
 
 export class InteractionTracker {
+  // Allows to track several events triggered by the same user interaction in the right Zone.
+  private shouldCreateNewZone = true;
+  private currentZone: Zone = Zone.root;
+
   constructor() {
+    const interactionTracker: InteractionTracker = this;
 
     const runTask = Zone.prototype.runTask;
     Zone.prototype.runTask = function(
@@ -44,26 +49,38 @@ export class InteractionTracker {
       const time = Date.now();
 
       console.warn('Running task');
+      console.log(task);
       console.log(task.zone);
 
       let taskZone = this;
       if (isTrackedElement(task)) {
         console.log('Click detected');
 
-        const traceId = randomTraceId();
-        const tracingZone = Zone.root.fork({
-          name: traceId,
-          properties: {
-            isTracingZone: true,
-            traceId,
-          },
-        });
+        if (interactionTracker.shouldCreateNewZone) {
+          interactionTracker.shouldCreateNewZone = false;
+
+          // Timeout of 50 ms to reset flags to allow the creation of a new zone for a
+          // new user interaction.
+          setTimeout(() => {
+            interactionTracker.shouldCreateNewZone = true;
+            interactionTracker.currentZone = Zone.root;
+          }, 50);
+
+          const traceId = randomTraceId();
+          interactionTracker.currentZone = Zone.root.fork({
+            name: traceId,
+            properties: {
+              isTracingZone: true,
+              traceId,
+            },
+          });
+          console.log('New zone:');
+          console.log(interactionTracker.currentZone);
+        }
 
         // Change the zone task.
-        task._zone = tracingZone;
-        taskZone = tracingZone;
-        console.log('New zone:');
-        console.log(taskZone);
+        task._zone = interactionTracker.currentZone;
+        taskZone = interactionTracker.currentZone;
       } else {
         // If we already are in a tracing zone, just run the task in our tracing zone.
         if (task.zone && task.zone.get('isTracingZone')) {
--- a/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
@@ -34,10 +34,13 @@ export type AsyncTask = Task & {
 
 export class InteractionTracker {
   // Allows to track several events triggered by the same user interaction in the right Zone.
-  private shouldCreateNewZone = true;
-  private currentZone: Zone = Zone.root;
+  private currentEventTracingZone?: Zone = undefined;
+
+  // Delay of 50 ms to reset currentEventTracingZone.
+  private readonly RESET_TRACING_ZONE_DELAY: number = 50;
 
   constructor() {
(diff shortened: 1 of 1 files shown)


packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -33,7 +33,12 @@ export type AsyncTask = Task & {
 };
 
 export class InteractionTracker {
+  // Allows to track several events triggered by the same user interaction in the right Zone.
+  private shouldCreateNewZone = true;
+  private currentZone: Zone = Zone.root;
+
   constructor() {
+    const interactionTracker: InteractionTracker = this;
```
draffensperger: Why do we need to capture `this` in a variable?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -33,7 +33,12 @@ export type AsyncTask = Task & {
 };
 
 export class InteractionTracker {
+  // Allows to track several events triggered by the same user interaction in the right Zone.
+  private shouldCreateNewZone = true;
+  private currentZone: Zone = Zone.root;
```
draffensperger: What would you think about calling this something like `currentEventTracingZone` or similar? When I see `currentZone` it makes me wonder how that is different from `Zone.current`.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -33,7 +33,12 @@ export type AsyncTask = Task & {
 };
 
 export class InteractionTracker {
+  // Allows to track several events triggered by the same user interaction in the right Zone.
+  private shouldCreateNewZone = true;
```
draffensperger: Rather than keeping a separate boolean for this, what would you think about just making the current event tracing zone be optional and resetting it to `undefined` in the timer?

So something like `private currentEventTracingZone?: Zone`

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -44,26 +49,38 @@ export class InteractionTracker {
...
-            isTracingZone: true,
-            traceId,
-          },
-        });
+        if (interactionTracker.shouldCreateNewZone) {
+          interactionTracker.shouldCreateNewZone = false;
+
+          // Timeout of 50 ms to reset flags to allow the creation of a new zone for a
+          // new user interaction.
+          setTimeout(() => {
```
draffensperger: Could you explicitly run this in the root zone just in case this happens to be in a deeper zone? (E.g. the Angular zone when done in Angular)

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -44,26 +49,38 @@ export class InteractionTracker {
...
+            interactionTracker.shouldCreateNewZone = true;
+            interactionTracker.currentZone = Zone.root;
+          }, 50);
```
draffensperger: Could you make the `50` a constant?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: This is necessary for monkey-patched methods (e.g. runTask) as the scope of these change we need to keep track of the interaction tracker.

draffensperger: Can you try using arrow functions `() => ` rather than a `function` declaration? 

Arrow functions capture the `this` variable automatically, see https://hackernoon.com/javascript-es6-arrow-functions-and-lexical-this-f2a3e2a5e8c4

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -34,10 +34,13 @@ export type AsyncTask = Task & {
 
 export class InteractionTracker {
   // Allows to track several events triggered by the same user interaction in the right Zone.
-  private shouldCreateNewZone = true;
-  private currentZone: Zone = Zone.root;
+  private currentEventTracingZone?: Zone = undefined;
+
+  // Delay of 50 ms to reset currentEventTracingZone.
+  private readonly RESET_TRACING_ZONE_DELAY: number = 50;
```
draffensperger: Nit: can you make this a top level constant rather than a field of the class? You can just use `const RESET_...` at the top of the file

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -32,43 +32,60 @@ export type AsyncTask = Task & {
   _zone: Zone;
 };
 
+// Delay of 50 ms to reset currentEventTracingZone.
+const RESET_TRACING_ZONE_DELAY = 50;
+
 export class InteractionTracker {
-  constructor() {
+  // Allows to track several events triggered by the same user interaction in the right Zone.
+  private currentEventTracingZone?: Zone = undefined;
```
draffensperger: Could you remove the `= undefined;` part since that's implied in TypeScript?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -32,43 +32,60 @@ export type AsyncTask = Task & {
...
-    ) {
+    ) => {
       const time = Date.now();
 
       console.warn('Running task');
+      console.log(task);
       console.log(task.zone);
 
-      let taskZone = this;
+      let taskZone: Zone = Zone.current;
```
draffensperger: The `: Zone` is implied by the assignment. What would you think about removing it?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -32,43 +32,60 @@ export type AsyncTask = Task & {
...
-        const traceId = randomTraceId();
-        const tracingZone = Zone.root.fork({
-          name: traceId,
-          properties: {
-            isTracingZone: true,
-            traceId,
-          },
-        });
+        if (this.currentEventTracingZone === undefined) {          
+
```
draffensperger: Nit: could you remove this extra blank line?

PR #89, opened 2019-06-05
Detect when an interaction is stable and measure total time to complete the interaction.

crdgonzalezca:
This adds the functionality to detect when an interaction is stable and measure the time to stable. This only takes in count the micro/macro tasks to count the tasks in the interaction.

### DIFF
--- a/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
@@ -52,7 +52,7 @@ export class InteractionTracker {
       console.log(task);
       console.log(task.zone);
 
-      const interceptingElement = isTrackedElement(task);
+      const interceptingElement = getTrackedElement(task);
       let taskZone = Zone.current;
       if (interceptingElement) {
         console.log('Click detected');
@@ -152,7 +152,9 @@ export class InteractionTracker {
     stopWatch.incrementTaskCount();
 
     if (interactionId in this.interactionCompletionTimeouts) {
-      // This may get called a lot. Run in the root zone.
+      // Clear the task that is supposed to complete the interaction as there are new
+      // tasks incrementing the task cout. Sometimes the task count might be 0
+      // but the interaction has more scheduled tasks.
       Zone.root.run(() => {
         clearTimeout(this.interactionCompletionTimeouts[interactionId]);
         delete this.interactionCompletionTimeouts[interactionId];
@@ -190,7 +192,8 @@ export class InteractionTracker {
 
     if (this.interactionCompletionTimeouts[interactionId] !== undefined) return;
 
-    // This may get called a lot. Run in the root zone.
+    // Add a task to the queue that will actually complete the interaction in case
+    // there are no more scheduled tasks ahead it.
     Zone.root.run(() => {
       this.interactionCompletionTimeouts[interactionId] = setTimeout(() => {
         this.completeInteraction(interactionId);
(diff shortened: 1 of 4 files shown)


packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -15,48 +15,48 @@
...
       applyThis: unknown,
       applyArgs: unknown
     ) => {
-      const time = Date.now();
-
       console.warn('Running task');
       console.log(task);
       console.log(task.zone);
 
+      const interceptingElement = isTrackedElement(task);
```
draffensperger: Since the function is returning the element now, what would you think about calling it `getTrackedElement`?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -90,43 +96,140 @@ export class InteractionTracker {
...
+  /** Increments the count of outstanding tasks for a given interaction id. */
+  private incrementTaskCount(interactionId: string) {
+    const stopWatch = this.getStopwatch(interactionId);
+    if (!stopWatch) return;
+    stopWatch.incrementTaskCount();
+
+    if (interactionId in this.interactionCompletionTimeouts) {
+      // This may get called a lot. Run in the root zone.
+      Zone.root.run(() => {
+        clearTimeout(this.interactionCompletionTimeouts[interactionId]);
```
draffensperger: What's the role of this interaction completion timeout? Why isn't it sufficient to just count tasks as they are scheduled/completed/cancelled?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -90,43 +96,140 @@ export class InteractionTracker {
...
+  return !!(task.zone && task.zone.get('isTracingZone'));
+}
+
+/**
+ * Whether or not a task should be tracked as part of an interaction.
+ */
+function shouldCountTask(task: Task): boolean {
+  if (!task.data) return false;
+
+  // Don't count periodic tasks like setInterval as they will be repeatedly
```
draffensperger: Why not count these as long as the timeout is < 1 second?

packages/opencensus-web-instrumentation-zone/src/on-page-interaction.ts
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { OnPageInteractionData } from './zone-types';
+
+/** A helper class for tracking on page interactions. */
+export class OnPageInteractionStopwatch {
+  private taskCount = 0;
+  private readonly startTimeMs = Date.now();
```
draffensperger: Could you use `performance.now` for higher accuracy?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: We need this to instead of declaring an interaction to be complete when the number of active interactions reaches 0 we add a task to the queue that will actually complete the interaction if none of the tasks scheduled ahead of it try and increment the task counter for the given interaction id.

Sometimes the count of tasks might be 0 but the interaction has more tasks.

draffensperger: OK, could you add a comment to explain all that?

PR #98, opened 2019-06-12
Send user interaction trace

crdgonzalezca:
This PR creates the a new RootSpan when a new user interaction is detected. This new root span will be associated to a new Zone. The root span duration represents the time to complete the user interaction.

The root span's name will be taken following some previously established rules to name the span. For this case, the example uses the `data-ocweb-id` property. 

* The span also contains some attributes like tag name of element triggering the user interaction, event that triggered it and URL.
*Changed the Tracing class as Singleton as we want only one instance of this.
*Update tracer tests.

Here an example of a new trace containing the root span.

image

### DIFF
--- a/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
@@ -61,44 +61,10 @@ export class InteractionTracker {
       let taskZone = Zone.current;
       if (interceptingElement) {
         console.log('Click detected');
+
         if (this.currentEventTracingZone === undefined) {
-          const traceId = randomTraceId();
-          const spanOptions = {
-            name: resolveInteractionName(interceptingElement, task.eventName),
-            spanContext: {
-              traceId,
-              // This becomes the parentSpanId field of the root span, and the actual
-              // span ID for the root span gets assigned to a random number.
-              spanId: '',
-            },
-            kind: SpanKind.UNSPECIFIED,
-          };
-          // Start a new RootSpan for a new user interaction.
-          tracing.tracer.startRootSpan(spanOptions, root => {
-            // As startRootSpan creates the zone and Zone.current corresponds to the
-            // new zone, we have to set the currentEventTracingZone with the Zone.current
-            // to capture the new zone.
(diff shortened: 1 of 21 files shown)


packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -220,7 +253,43 @@ function getTrackedElement(task: AsyncTask): HTMLElement | null {
...
+): string {
+  if (!element.getAttribute) return '';
+  if (element.hasAttribute('disabled')) {
+    return '';
+  }
+  let interactionName = element.getAttribute('data-ocweb-id');
+  if (!interactionName) {
+    const elementId = element.getAttribute('id')
+      ? element.getAttribute('id')
+      : '';
```
johnbryan: can abbreviate to: = element.getAttribute('id) || '';

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -85,10 +100,12 @@ export class InteractionTracker {
         }
 
         // Change the zone task.
-        task._zone = this.currentEventTracingZone;
-        taskZone = this.currentEventTracingZone;
-        this.incrementTaskCount(task.zone.get('traceId'));
-      } else if (task.zone && task.zone.get('isTracingZone')) {
+        if (this.currentEventTracingZone) {
```
johnbryan: can this just be "else"

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -58,18 +63,28 @@ export class InteractionTracker {
         console.log('Click detected');
         if (this.currentEventTracingZone === undefined) {
           const traceId = randomTraceId();
-          this.currentEventTracingZone = Zone.root.fork({
-            name: traceId,
-            properties: {
-              isTracingZone: true,
+          const spanOptions = {
```
johnbryan: the Zone.prototype.runtask() function is getting pretty long. If you see a way to cleanly extract a helper function that help keep it clear.

PR #102, opened 2019-06-20
Detect route transitions and send 'Navigation' traces.

crdgonzalezca:
Monkey-patches `History API` to detect a route change and then maybe change the name of the interaction following the 'Navigation' naming approach.
Also, update the user interaction example to add a route transitions.

### DIFF
--- a/packages/opencensus-web-core/src/trace/model/tracer.ts
+++ b/packages/opencensus-web-core/src/trace/model/tracer.ts
@@ -69,10 +69,8 @@ export class Tracer extends TracerBase implements webTypes.Tracer {
         },
       },
     };
-    // Start the new zone from the current, for that, we assume
-    // all click handlers have the same parent zone.
-    const newZone = Zone.current.fork(zoneSpec);
 
+    const newZone = Zone.current.fork(zoneSpec);
     return newZone.run(() => {
       super.startRootSpan(options, root => {
         // Set the currentRootSpan to the new created root span.
--- a/packages/opencensus-web-core/src/trace/model/tracer.ts
+++ b/packages/opencensus-web-core/src/trace/model/tracer.ts
@@ -69,7 +69,9 @@ export class Tracer extends TracerBase implements webTypes.Tracer {
         },
       },
     };
-    const newZone = Zone.root.fork(zoneSpec);
+    // Start the new zone from the current, for that, we assume
+    // all click handlers have the same parent zone.
+    const newZone = Zone.current.fork(zoneSpec);
 
     return newZone.run(() => {
       super.startRootSpan(options, root => {
--- a/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
@@ -68,28 +68,22 @@ export class InteractionTracker {
         interceptingElement,
         task.eventName
       );
-      let taskZone = Zone.root;
       if (interceptingElement) {
         if (this.currentEventTracingZone === undefined && interactionName) {
           this.startNewInteraction(
             interceptingElement,
             task.eventName,
+            task.zone,
             interactionName
           );
         }
         if (this.currentEventTracingZone) {
           task._zone = this.currentEventTracingZone;
-          taskZone = this.currentEventTracingZone;
-        } else {
-          task._zone = Zone.root;
         }
         this.incrementTaskCount(getTraceId(task.zone));
-      } else if (isTrackedTask(task)) {
-        // If we already are in a tracing zone, just run the task in our tracing zone.
-        taskZone = task.zone;
       }
       try {
-        return runTask.call(taskZone as {}, task, applyThis, applyArgs);
+        return runTask.call(task.zone as {}, task, applyThis, applyArgs);
       } finally {
         console.log('Run task finished.');
         if (
@@ -140,6 +134,7 @@ export class InteractionTracker {
   private startNewInteraction(
     interceptingElement: HTMLElement,
     eventName: string,
+    taskZone: Zone,
     interactionName: string
   ) {
     const traceId = randomTraceId();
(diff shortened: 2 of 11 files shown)


packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -61,15 +68,16 @@ export class InteractionTracker {
         interceptingElement,
         task.eventName
       );
-      let taskZone = Zone.current;
+      let taskZone = Zone.root;
```
draffensperger: Why is this changing?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -237,6 +242,47 @@ export class InteractionTracker {
...
+  // but if there is an user interaction that triggers a route transition
+  // while those interactions are still in tracking, only that interaction
+  // will have a `Navigation` name. Otherwise, if this is not patched, the
+  // other interactions will change the name to `Navigation` even if they
+  // did not cause the route transition.
+  private patchHistoryApi() {
+    const pushState = history.pushState;
+    history.pushState = (
+      //tslint:disable:no-any
+      data: any,
```
draffensperger: Can you use `unknown` here instead of `any` and so remove the `tslint` comment? Same below as well.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -237,6 +242,47 @@ export class InteractionTracker {
...
+
+  // Monkey-patch `History API` to detect route transitions.
+  // This is necessary because there might be some cases when
+  // there are several interactions being tracked at the same time
+  // but if there is an user interaction that triggers a route transition
+  // while those interactions are still in tracking, only that interaction
+  // will have a `Navigation` name. Otherwise, if this is not patched, the
+  // other interactions will change the name to `Navigation` even if they
+  // did not cause the route transition.
+  private patchHistoryApi() {
```
draffensperger: Do you also need to monkey-patch the `back`, `forward`, and `go` methods? See https://developer.mozilla.org/en-US/docs/Web/API/History

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -237,6 +242,47 @@ export class InteractionTracker {
...
+      title: string,
+      url?: string | null | undefined
+    ) => {
+      replaceState.call(history, data, title, url);
+      this.maybeUpdateInteractionName();
+    };
+  }
+
+  private maybeUpdateInteractionName() {
+    console.log('location changed!');
```
draffensperger: Can this be removed?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -237,6 +242,47 @@ export class InteractionTracker {
...
+    // Route transition is detected, thus the current interaction name might change.
+    const rootSpan = getRootSpan(Zone.current);
```
draffensperger: Can you get this from the tracer instance instead?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -237,6 +242,47 @@ export class InteractionTracker {
...
+    if (rootSpan && rootSpan.name.startsWith('<')) {
```
draffensperger: What does the `rootSpan.name.startsWith('<')` check for? Is that in case the user gave an explicit tag for the element?

That's an interesting question as to whether the navigation name or the element tag name "wins". At any rate, I think commenting to explain it would be good.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -237,6 +242,47 @@ export class InteractionTracker {
...
+      rootSpan.name = 'Navigation ' + location.pathname;
```
draffensperger: Would it make sense to double check that `location.pathname` actually changed since the interaction started (you are still capturing it above, though as an `href` so I think if you just changed that to ? That way you can double check that the user didn't call `replaceState` and just ended up going to a URL they were already at.

I also think that doing `${location.pathname}#${location.hash}` (and maybe putting that in a function) would be good because sometimes people do routing just based on the hash in the URL.

packages/opencensus-web-instrumentation-zone/src/on-page-interaction.ts
```diff
@@ -47,7 +47,7 @@ export class OnPageInteractionStopwatch {
     const rootSpan = this.data.rootSpan;
     rootSpan.addAttribute('EventType', this.data.eventType);
     rootSpan.addAttribute('TargetElement', this.data.target.tagName);
-    rootSpan.addAttribute(ATTRIBUTE_HTTP_URL, location.href);
+    rootSpan.addAttribute(ATTRIBUTE_HTTP_URL, this.data.originLocation);
```
draffensperger: Smart! Seems you do need to keep capturing `href` in the data for each interaction zone in addition to path and hash I think.

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -31,6 +31,7 @@ export type AsyncTask = Task & {
 
 /** Data used to create a new OnPageInteractionStopwatch. */
 export interface OnPageInteractionData {
+  originLocation: string;
```
draffensperger: What would you think about calling this `startLocationHref` or similar? When I hear "origin" I think of the origin (host) of the site, and I think adding the `Href` at the end is nice in case you add things like path and hash too.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: In #100 I update tracking rules and changed it to track only the interaction that can be labeled, that means that interactions related to clicks on the document are not tracked, the I found our this was a bug so the best initial value for taskZone is Zone.root

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: Folowing [the browser compatibility](https://developer.mozilla.org/en-US/docs/Web/API/History#Browser_compatibility) I see it is not necessary to change the name using the hash as it does not actually do a route transition.

Also, chatting offline, we found valuable to store the `location.pathname` in the transition.

crdgonzalezca:
Offline chat: Change the way the zones are created, now the zones are `fork`ed from the current task zone, that way all new zones have the same parent.

packages/opencensus-web-core/src/trace/model/tracer.ts
```diff
@@ -69,7 +69,9 @@ export class Tracer extends TracerBase implements webTypes.Tracer {
         },
       },
     };
-    const newZone = Zone.root.fork(zoneSpec);
+    // Start the new zone from the current, for that, we assume
```
draffensperger: Could you move this comment to the interaction specific code?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -153,20 +148,24 @@ export class InteractionTracker {
...
       kind: SpanKind.UNSPECIFIED,
     };
-    // Start a new RootSpan for a new user interaction.
-    tracing.tracer.startRootSpan(spanOptions, root => {
-      // As startRootSpan creates the zone and Zone.current corresponds to the
-      // new zone, we have to set the currentEventTracingZone with the Zone.current
-      // to capture the new zone.
-      this.currentEventTracingZone = Zone.current;
+    // Start a new RootSpan for a new user interaction, also, creates the new zone
+    // from the coming task.zone.
```
draffensperger: What would you think about re-working `from the coming task.zone` to `forked from the original task.zone`?

PR #103, opened 2019-06-24
Refactor Tracer to check presence of Zone global variable

crdgonzalezca:
This checks the presence of `Zone` in Tracer as there might be some case where `zone.js` library is not imported.
Also, modify `Tracer` to be Singleton as there will be only one instance.

### DIFF
--- a/packages/opencensus-web-core/src/trace/model/tracer.ts
+++ b/packages/opencensus-web-core/src/trace/model/tracer.ts
@@ -22,62 +22,85 @@ import { randomTraceId } from '../../common/id-util';
 
 /** Tracer manages the current root span and trace header propagation. */
 export class Tracer extends TracerBase implements webTypes.Tracer {
+  private static singletonInstance: Tracer;
+
+  /** Gets the tracer instance. */
+  static get instance(): Tracer {
+    return this.singletonInstance || (this.singletonInstance = new this());
+  }
+
+  // Variable to store current root span in case the Zone global variable is not present.
+  // For that case we only need to store only one current root span.
+  private rootSpan = new RootSpan(this);
+
   /**
    * Gets the current root span associated to Zone.current.
-   * If the current zone does not have a root span (e.g. root zone),
-   * create a new root.
+   * If the current zone does not have a root span (e.g. root zone) or
+   * `Zone` is not present return the value store in the unique root span.
    */
   get currentRootSpan(): Span {
-    if (Zone.current.get('data')) {
+    if (this.isZonePresent() && Zone.current.get('data')) {
       return Zone.current.get('data').rootSpan;
     }
-    return new RootSpan(this);
+    return this.rootSpan;
   }
 
   /**
    * Sets the current root span to the current Zone.
    * If the current zone does not have a 'data' property (e.g. root zone)
-   * do not set the root span.
+   * or `Zone` is not present, just assign the root span to a variable.
    */
   set currentRootSpan(root: Span) {
-    if (Zone.current.get('data')) {
+    if (this.isZonePresent() && Zone.current.get('data')) {
       Zone.current.get('data')['rootSpan'] = root;
+    } else {
+      this.rootSpan = root as RootSpan;
     }
   }
 
   /**
-   * Creates a new Zone and start a new RootSpan to `currentRootSpan` associating
-   * the new RootSpan to the new Zone. Thus, there might be several root spans
-   * at the same time.
+   * Creates a new Zone (in case `Zone` global variable is present) and start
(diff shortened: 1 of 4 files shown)


packages/opencensus-web-core/src/trace/model/tracer.ts
```diff
@@ -22,62 +22,85 @@ import { randomTraceId } from '../../common/id-util';
 
 /** Tracer manages the current root span and trace header propagation. */
 export class Tracer extends TracerBase implements webTypes.Tracer {
+  private static singletonInstance: Tracer;
+
+  /** Gets the tracer instance. */
+  static get instance(): Tracer {
```
draffensperger: What's the motivation for adding this `instance` getter?

packages/opencensus-web-core/src/trace/model/tracer.ts
```diff
@@ -22,62 +22,85 @@ import { randomTraceId } from '../../common/id-util';
...
+    return this.singletonInstance || (this.singletonInstance = new this());
+  }
+
+  // Variable to store current root span in case the Zone global variable is not present.
+  // For that case we only need to store only one current root span.
+  private rootSpan = new RootSpan(this);
```
draffensperger: Could we call this something like `rootSpanNoZone` or similar to indicate that it's only used if there is no Zone available?

packages/opencensus-web-core/src/trace/model/tracer.ts
```diff
@@ -22,62 +22,85 @@ import { randomTraceId } from '../../common/id-util';
...
-      super.startRootSpan(options, root => {
+      const newZone = Zone.current.fork(zoneSpec);
+      return newZone.run(() => {
+        super.startRootSpan(options, root => {
+          // Set the currentRootSpan to the new created root span.
+          this.currentRootSpan = root;
+          return fn(root);
+        });
+      });
+    } else {
```
draffensperger: Do you need the `else ` here since there is a `return` above?

packages/opencensus-web-core/src/trace/model/tracer.ts
```diff
@@ -108,4 +131,9 @@ export class Tracer extends TracerBase implements webTypes.Tracer {
 
   /** Binds trace context to NodeJS event emitter. No-op for opencensus-web. */
   wrapEmitter(emitter: webTypes.NodeJsEventEmitter) {}
+
+  isZonePresent(): boolean {
+    //tslint:disable:no-any
```
draffensperger: Rather than using `any`, could you create a new interface to cast `window` to?

For example:
```
interface WindowWithZone {
   Zone?: Function
}
```

packages/opencensus-web-core/src/trace/model/tracing.ts
```diff
@@ -23,7 +23,7 @@ export const NOOP_EXPORTER = new NoopExporter();
 /** Main interface for tracing. */
 export class Tracing implements webTypes.Tracing {
   /** Object responsible for managing a trace. */
-  readonly tracer = new Tracer();
+  readonly tracer = Tracer.instance;
```
draffensperger: Is this similar to how OpenCensus Node does it?

packages/opencensus-web-core/test/test-tracer.ts
```diff
@@ -17,10 +17,16 @@
 import * as webTypes from '@opencensus/web-types';
 import { Tracer } from '../src/trace/model/tracer';
 
+export interface WindowWithZone extends Window {
+  Zone?: Zone;
```
draffensperger: I'm not sure the type here should be `Zone`, since that would be the type of an individual zone and not the type of the Zone class itself. I think you could use `Function`, which is vague but true.

packages/opencensus-web-core/src/trace/model/tracer.ts
crdgonzalezca: I saw that opencensus node does it and I think is more understandable that `Tracer` is a Singleton class.

packages/opencensus-web-core/src/trace/model/tracing.ts
crdgonzalezca: Yes, I saw they do it like this, and it is more understandable that this class is Singleton.

packages/opencensus-web-core/src/trace/model/tracer.ts
draffensperger: OK, makes sense.

packages/opencensus-web-core/src/trace/model/tracing.ts
draffensperger: Sounds good!

PR #104, opened 2019-06-27
Add unit/integration testing for @opencensus/web-instrumentation-zone

crdgonzalezca:
- Add unit testing for `OnPageInteraction` class.
- Integration testing on InteractionTracker which creates fake interaction in other to test all kind of supported tasks (e.g. `setTimeout`, `Promise`, HTTP requests).

Additionally, there are some changes little changes:
- Change `InteractionTracker` as Singleton as there will be only one InteractionTracker instance always.
- (Fix) Reset `currentEventTracingZone` when the interaction has finished before it is reset automatically.
- Register a initial `SpanEventListener` in the `Tracing` (for this case the `NOOP_EXPORTER`) as this is helpful for testing.
- Stop tracking periodic tasks (`setInterval`) as they might never stop and interaction wouldn't be stable.


packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -375,10 +394,9 @@ function resolveInteractionName(
 function shouldCountTask(task: Task): boolean {
   if (!task.data) return false;
 
-  // Don't count periodic tasks with a delay greater than 1 s.
-  if (task.data.isPeriodic && (task.data.delay && task.data.delay >= 1000)) {
-    return false;
-  }
+  // Don't count periodic tasks like setInterval as they will be repeatedly
+  // called.
+  if (task.data.isPeriodic) return false;
```
draffensperger: Why did you change this to ignore all periodic tasks? I would think that calls to `setInterval` could be pretty relevant to the work of user interactions.

One other thing that should be a TODO for later I think - if you have a periodic task with > 1 second of time, then I think it should be executed outside of the tracing zone. Otherwise, if it ends up doing some work in the future it will still be associated with that same older tracing zone.

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+} from '../src/interaction-tracker';
+
+describe('InteractionTracker', () => {
+  const interactionTracker = InteractionTracker.instance;
+  let incrementTaskCountSpy: jasmine.Spy;
+  let decrementTaskCountSpy: jasmine.Spy;
+  let onEndSpanSpy: jasmine.Spy;
+
+  beforeEach(() => {
+    interactionTracker.resetCurrentTracingZone();
```
draffensperger: Why is this needed?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+    onEndSpanSpy.and.callFake((rootSpan: Span) => {
+      // Counts related to the click event.
+      expect(incrementTaskCountSpy.calls.count()).toBe(1);
+      expect(decrementTaskCountSpy.calls.count()).toBe(1);
+      expect(rootSpan.name).toBe('test interaction');
+      expect(rootSpan.ended).toBeTruthy();
+      // As there is another setTimeOut that completes the interaction, the
+      // span duraction is not precise, then only test if the interaction duration
+      // finishes within a range.
+      expect(rootSpan.duration).toBeLessThan(5);
```
draffensperger: Should you check a minimum time here too?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+      setTimeout(() => {
+        expect(incrementTaskCountSpy.calls.count()).toBe(2);
+        expect(decrementTaskCountSpy.calls.count()).toBe(1);
+      });
+    };
+
+    button.removeAttribute('disabled');
+    fakeInteraction(onclick2, button);
+    onEndSpanSpy.and.callFake((rootSpan: Span) => {
+      // There should be only 2 counted tasks related to the second interaction.
```
draffensperger: Could you also test this by making sure that `onEndSpanSpy` is only called once?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+      expect(incrementTaskCountSpy.calls.count()).toBe(1);
+      expect(decrementTaskCountSpy.calls.count()).toBe(1);
+      expect(rootSpan.name).toBe('test interaction');
+      expect(rootSpan.ended).toBeTruthy();
+      expect(rootSpan.duration).toBeLessThanOrEqual(10);
+      done();
+    });
+  });
+
+  it('should not create a new interaction when second click occurs before 50 ms.', done => {
```
draffensperger: Should this say something like "when second click *handler* occurs before 50 ms."? (May need to reword a bit to keep < 80 chars)

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+    // Create a button without 'data-ocweb-id' attribute.
+    const button = createButton('');
+    fakeInteraction(onclick, button);
+
+    onEndSpanSpy.and.callFake((rootSpan: Span) => {
+      expect(incrementTaskCountSpy.calls.count()).toBe(2);
+      expect(decrementTaskCountSpy.calls.count()).toBe(2);
+      expect(rootSpan.name).toBe('Navigation /test_navigation');
+      expect(rootSpan.ended).toBeTruthy();
+      expect(rootSpan.duration).toBeLessThanOrEqual(10);
```
draffensperger: What if there was a constant like TIME_BUFFER or similar that we use here and other places to indicate that we just expect that these interactions take a little extra time to complete due to the `setTimeout` that is needed for the final completion (and have a comment that explains that)?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+      expect(decrementTaskCountSpy.calls.count()).toBe(0);
+      const promise = getPromise();
+      promise.then(() => {
+        expect(incrementTaskCountSpy.calls.count()).toBe(2);
+        expect(decrementTaskCountSpy.calls.count()).toBe(1);
+        setTimeout(() => {
+          expect(incrementTaskCountSpy.calls.count()).toBe(3);
+          expect(decrementTaskCountSpy.calls.count()).toBe(2);
+          doHTTPRequest();
+        }, 1);
```
draffensperger: Should this time be longer so that the total is measurable in the test?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -0,0 +1,408 @@
...
+  function getPromise() {
+    return new Promise(resolve => {
+      resolve();
+    });
+  }
+
+  function doHTTPRequest() {
+    const xhr = new XMLHttpRequest();
+    spyOn(xhr, 'send').and.callFake(() => {
+      setTimeout(noop, 60);
```
draffensperger: What if this `60` was a constant like XHR_TIME? And what if we also had a `SET_TIMEOUT_TIME`? Then in the tests you could have something that adds up the times so something like this:

```
const interactionTime = XHR_TIME + SET_TIMEOUT_TIME;
expect(rootSpan.duration).toBeGreaterThanOrEqual(interactionTime);
expect(rootSpan.duration).toBeLessThanOrEqual(interactionTime + BUFFER_TIME);
```

packages/opencensus-web-instrumentation-zone/test/test-on-page-interaction.ts
```diff
@@ -0,0 +1,85 @@
...
+        target,
+        rootSpan: root,
+      };
+      interaction = new OnPageInteractionStopwatch(interactionData);
+    });
+    it('should stop an on page interaction and end the root span', () => {
+      spyOn(tracer, 'onEndSpan');
+      interaction.stopAndRecord();
+
+      expect(root.attributes['EventType']).toBe('click');
```
draffensperger: Should the test cases above for the fake user interactions also test these attributes?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: I found out that if we track periodic tasks, the interaction might not finish as they are called repeatedly, so we won't be able to determine the stability of the interaction.

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
crdgonzalezca: Not longer needed as now the `InteractionTracker` does this every time the interaction finishes.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -30,11 +30,15 @@ import {
 declare const Zone: ZoneType & { prototype: Zone };
 
 // Delay of 50 ms to reset currentEventTracingZone.
-const RESET_TRACING_ZONE_DELAY = 50;
+export const RESET_TRACING_ZONE_DELAY = 50;
 
 export class InteractionTracker {
   // Allows to track several events triggered by the same user interaction in the right Zone.
-  private currentEventTracingZone?: Zone;
+  currentEventTracingZone?: Zone;
```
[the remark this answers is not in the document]
crdgonzalezca: It's necessary to test when there are overlapping interactions and check the interactions have different ID. Also, when there are several event clicks corresponding to the same interaction, we want to make sure there is not a new zone for those click events. I am open to receive recommendations on a way to avoid exposing this.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -170,18 +176,23 @@ export class InteractionTracker {
...
+  }
+
+  resetCurrentTracingZone() {
+    this.currentEventTracingZone = undefined;
+    this.currentResetTracingZoneTimeout = undefined;
   }
 
   /** Increments the count of outstanding tasks for a given interaction id. */
-  private incrementTaskCount(interactionId: string) {
+  incrementTaskCount(interactionId: string) {
```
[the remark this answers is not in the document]
crdgonzalezca: The motivation is to make sure we actually count the correct amount of tasks involved in the interaction. The duration of interaction is a good test but might give a wrong testing on the amount of tracked tasks for every interaction, as we actually want to check that, for example, macrotasks are handled. I am open to different approach though.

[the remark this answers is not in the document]
draffensperger: Well, I guess to me the fact that it is counting tasks is more of an implementation detail that is not exposed to the end user of the tracker. So for an integration level test, I think we should be testing the surface as experienced by the user as close as possible.

Another idea would be to try to split out the part of the code that does the task counting and test that separately more as an implementation detail to make sure the counts are correct. And then in the integration test to just measure the total time and span name/metadata.

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
crdgonzalezca: In this case, this noop sync function does not have any logic.Then, the minimum time would be 0. I think that just checking the max time is fine.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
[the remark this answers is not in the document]
crdgonzalezca: Update: I stopped exposing this as final decision was to leave this kind of testing of a future refactoring on `InteractionTracker`

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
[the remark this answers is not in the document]
crdgonzalezca: Final decision: I decided to stop testing the task count as it should be more an implementation testing more than integration testing. Also, this could be done in a future refactoring on `InteractinTracker` to test every specific module.

packages/opencensus-web-instrumentation-zone/test/test-on-page-interaction.ts
crdgonzalezca: Thanks, Added.

packages/opencensus-web-core/src/trace/model/tracing.ts
```diff
@@ -39,6 +39,12 @@ export class Tracing implements webTypes.Tracing {
     return this.singletonInstance || (this.singletonInstance = new this());
   }
 
+  constructor() {
+    // register the noop exporter to have an span event listener,
+    // helpful for testing.
+    this.tracer.registerSpanEventListener(this.exporter);
```
draffensperger: Could you do this in the tests rather than here in the production code?

packages/opencensus-web-instrumentation-zone/test/test-on-page-interaction.ts
```diff
@@ -0,0 +1,85 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { RootSpan, Tracer } from '@opencensus/web-core';
+import { OnPageInteractionStopwatch } from '../src/on-page-interaction';
+
+describe('OnPageInteraction', () => {
```
draffensperger: Should this be 'OnPageInteractionStopwatch'? And should the file be something like `test-on-page-interaction-stopwatch.ts`?

packages/opencensus-web-core/src/trace/model/tracing.ts
crdgonzalezca: Done

packages/opencensus-web-instrumentation-zone/test/test-on-page-interaction.ts
crdgonzalezca: Updated.

PR #118, opened 2019-07-09
Propagate trace context header

crdgonzalezca:
Intercept XHR tasks, add the `traceparent` context header. Also, creates a child span for every XHR task so that way the client side will have their spans. Final result is shown as follows:
image

In order to add some attributes to the child span like the `HTTP method`, the ` XMLHttpRequest.prototype.open` method is monkey patched.

### DIFF
--- a/examples/user_interaction/client/src/index.js
+++ b/examples/user_interaction/client/src/index.js
@@ -45,7 +45,7 @@ const routing = (
 ReactDOM.render(routing, document.getElementById('root'));
 
 window.ocAgent = 'http://localhost:55678';
-// For the purpose of this example, just match all hosts.
+// For the purpose of this example, send trace header to all hosts.
 window.ocTraceHeaderHostRegex = /.*/;
 window.ocSampleRate = 1.0; // Sample at 100% for test only. Default is 1/10000.
 
--- a/examples/user_interaction/client/src/index.js
+++ b/examples/user_interaction/client/src/index.js
@@ -45,7 +45,8 @@ const routing = (
 ReactDOM.render(routing, document.getElementById('root'));
 
 window.ocAgent = 'http://localhost:55678';
-window.ocTraceOrigins = 'http://localhost:3000';
+// For the purpose of this example, just match all hosts.
+window.ocTraceHeaderHostRegex = /.*/;
 window.ocSampleRate = 1.0; // Sample at 100% for test only. Default is 1/10000.
 
 // Send the root span and child spans for the initial page load to the
--- a/examples/user_interaction/client/src/index.js
+++ b/examples/user_interaction/client/src/index.js
@@ -45,6 +45,7 @@ const routing = (
 ReactDOM.render(routing, document.getElementById('root'));
 
 window.ocAgent = 'http://localhost:55678';
+window.ocTraceOrigins = 'http://localhost:3000';
 window.ocSampleRate = 1.0; // Sample at 100% for test only. Default is 1/10000.
 
 // Send the root span and child spans for the initial page load to the
--- a/examples/user_interaction/server/server.js
+++ b/examples/user_interaction/server/server.js
@@ -78,7 +78,8 @@ function handleRequest(request, response) {
 
     let result = '';
     let code = 200;
-    if(request.method !== 'GET'){
+    // Logic related to CORS pre-flight requests.
+    if(request.method === 'OPTIONS'){
       request.on('end', () => {
         span.end();
         response.statusCode = code;
--- a/examples/user_interaction/server/server.js
+++ b/examples/user_interaction/server/server.js
@@ -63,7 +63,7 @@ function calculatePrimeNumbers() {
 
 /** A function which handles requests and send response. */
 function handleRequest(request, response) {
-  const span = tracer.startChildSpan({ name: 'octutorials.handleRequest' });
+  const span = tracer.startChildSpan({ name: 'ocweb.handleRequest' });
 
   try {
     let body = [];
@@ -72,16 +72,24 @@ function handleRequest(request, response) {
 
     // Necessary headers because the Node.js and React dev servers run in different ports.
     response.setHeader('Access-Control-Allow-Origin', '*');
-    response.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS, PUT, PATCH, DELETE');
-    response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With,content-type');
+    response.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
+    response.setHeader('Access-Control-Allow-Headers', 'traceparent');
     response.setHeader('Access-Control-Allow-Credentials', true);
 
     let result = '';
     let code = 200;
+    if(request.method !== 'GET'){
+      request.on('end', () => {
+        span.end();
+        response.statusCode = code;
+        response.end();
+      });
+      return;
+    }
     if (url.parse(request.url).pathname === '/sleep') {
       console.log("Sleeping...")
       const time = Date.now();
-      sleep.sleep(5);
+      sleep.sleep(2);
       result = { time: Date.now() - time, value: "" };
       console.log("Finished.")
     } else if (url.parse(request.url).pathname === '/prime_numbers') {
@@ -109,7 +117,7 @@ function handleRequest(request, response) {
 function setupTracerAndExporters() {
   const zipkinOptions = {
     url: 'http://localhost:9411/api/v2/spans',
-    serviceName: 'opencensus_tutorial'
+    serviceName: 'opencensus_web_server'
   };
 
   // Creates Zipkin exporter
--- a/examples/user_interaction/server/server.js
+++ b/examples/user_interaction/server/server.js
@@ -70,16 +70,16 @@ function handleRequest(request, response) {
     request.on('error', err => console.log(err));
     request.on('data', chunk => body.push(chunk));
 
-    // Necessary headers because the Node.js and React dev servers run in different ports.
-    response.setHeader('Access-Control-Allow-Origin', '*');
-    response.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
-    response.setHeader('Access-Control-Allow-Headers', 'traceparent');
-    response.setHeader('Access-Control-Allow-Credentials', true);
-
+    // Necessary header because the Node.js and React dev servers run in different ports.
+    response.setHeader('Access-Control-Allow-Origin', '*');    
+    
     let result = '';
     let code = 200;
-    // Logic related to CORS pre-flight requests.
+    // Accept all CORS pre-flight requests
     if(request.method === 'OPTIONS'){
+      // Set headers to allow traceparent and subsequent get request.
+      response.setHeader('Access-Control-Allow-Headers', 'traceparent');
+      response.setHeader('Access-Control-Allow-Methods', 'GET');
       request.on('end', () => {
         span.end();
         response.statusCode = code;
--- a/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
@@ -59,7 +59,7 @@ export class InteractionTracker {
 
   // Map intended to keep track of current XHR objects
   // associated to a span.
-  private readonly xhrTasks = new Map<XHRWithUrl, Span>();
+  private readonly xhrSpans = new Map<XHRWithUrl, Span>();
 
   private static singletonInstance: InteractionTracker;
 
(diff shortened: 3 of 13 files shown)


examples/user_interaction/server/server.js
```diff
@@ -72,16 +72,24 @@ function handleRequest(request, response) {
...
     response.setHeader('Access-Control-Allow-Origin', '*');
-    response.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS, PUT, PATCH, DELETE');
-    response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With,content-type');
+    response.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
+    response.setHeader('Access-Control-Allow-Headers', 'traceparent');
     response.setHeader('Access-Control-Allow-Credentials', true);
 
     let result = '';
     let code = 200;
+    if(request.method !== 'GET'){
```
draffensperger: Is this logic for CORS preflght requests? If so, could you add a comment to explain that (and maybe match `request.method === 'OPTIONS'`)?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -50,6 +56,12 @@ export class InteractionTracker {
     [index: string]: number;
   } = {};
 
+  // Map intended to keep track of current XHR objects
+  // associated to a span.
+  private readonly xhrTasks: {
+    [index: string]: XMLHttpRequest;
```
draffensperger: Instead of `index` could you call this `spanId`? Could you also explain that we use object equality to match XHR instances to spans but that an object can only accept string or number keys so the mapping must be from span ID to XHR?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -186,6 +199,46 @@ export class InteractionTracker {
     });
   }
 
+  private interceptXHRTasks(task: AsyncTask) {
```
draffensperger: Nit: usually we name things that are acronyms in title case, so this would be `interceptXhrTasks`

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -186,6 +199,46 @@ export class InteractionTracker {
...
+    if (isTrackedTask(task) && task.target instanceof XMLHttpRequest) {
+      const rootSpan: RootSpan = task.zone.get('data').rootSpan;
```
draffensperger: What would you think about making both branches of this `if` statement into helper functions?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -186,6 +199,46 @@ export class InteractionTracker {
...
+      const xhr = task.target as XHRWithUrl;
+      if (xhr.readyState === XMLHttpRequest.OPENED) {
+        const childSpan = rootSpan.startChildSpan({
+          name: parseUrl(xhr.__zone_symbol__xhrURL).pathname,
```
draffensperger: Could you add a comment that `__zone_symbol__xhrURL` is set by the Zone monkey patch?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
draffensperger: An alternative would be to use a `Map`, which is only available in newer versions of JavaScript (old versions are polyfilled), but which allows setting objects as the keys.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -186,6 +199,46 @@ export class InteractionTracker {
...
+      } else if (xhr.readyState === XMLHttpRequest.DONE) {
+        const spanId =
+          Object.keys(this.xhrTasks).find(
+            spanId => this.xhrTasks[spanId] === xhr
+          ) || '';
+        const childSpan = rootSpan.spans.find(span => span.id === spanId);
+        if (childSpan) {
+          childSpan.addAttribute(
+            ATTRIBUTE_HTTP_STATUS_CODE,
+            xhr.status.toString()
```
draffensperger: Why do you use `toString` here? I think we can actually set the status as a number

packages/opencensus-web-instrumentation-zone/src/monkey-patching.ts
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+import { XHRWithUrl } from './zone-types';
+
+export function doPatching() {
+  patchXMLHttpRequestOpen();
+}
+
+// Patch the `XMLHttpRequest.open` method to add method used for the request
```
draffensperger: Could you document that this patch is needed because Zone.js does not capture the method from XHR the way that it captures URL as `__zone_symbol__xhrURL`?

packages/opencensus-web-instrumentation-zone/src/util.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { WindowWithOcwGlobals } from './zone-types';
+import { parseUrl } from '@opencensus/web-core';
+
+/** Check that the trace */
+export function traceOriginMatchesOrSameOrigin(): boolean {
+  const traceOrigin = (window as WindowWithOcwGlobals).ocTraceOrigins as string;
+  if (!traceOrigin) return false;
+  const parsedUrl = parseUrl(traceOrigin);
+  const regex = /.*/;
```
draffensperger: Shouldn't this regex come from the `ocTraceOrigins` field?

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -38,6 +38,13 @@ export interface OnPageInteractionData {
   rootSpan: RootSpan;
 }
 
+/** Allows monkey-patching XMLHttpRequest and to obtain the request URL. */
+export type XHRWithUrl = HTMLElement &
```
draffensperger: Why does this need the `HTMLElement` type here?

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -51,4 +58,7 @@ export declare interface WindowWithOcwGlobals extends Window {
    * specified sample rate. If not specified, a default sampling rate is used.
    */
   ocSampleRate?: number;
+
+  //
```
draffensperger: Could you add some documentation here to explain that this controls the regex of what origins will get the trace context header sent to them? Maybe also rename it to be a bit clearer there, e.g. called `ocTraceHeaderHostRegex` or similar?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: As the type `number` in Typescript is a double-precision 64-bit, if we send just the number, when the JSON object is created for this span, the type value will be `doublevalue` then the status code is not shown, but if a string is sent rather than a number it work fine.

draffensperger: Interesting, I think that's probably a bug in how it gets exported to Stackdriver, but for now I'm OK with making it a string. Could you add a TODO that we should investigate changing this to a number once we know that it can be exported correctly?

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
crdgonzalezca: `HTMLElement` is necessary when the xhr is captured from the tasks target as the Zone monkey-patch parses xhrs as `HTMLElement & XMLHttpRequest`. 
Added this as a comment also.

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: Done

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
crdgonzalezca: I changed this as `Map` as it's more readable and easier to understand.

examples/user_interaction/client/src/index.js
```diff
@@ -45,6 +45,8 @@ const routing = (
 ReactDOM.render(routing, document.getElementById('root'));
 
 window.ocAgent = 'http://localhost:55678';
+// For the purpose of this example, just match all hosts.
```
draffensperger: Nit: maybe say "For the purpose of this example, send trace header to all hosts"? (This will help people reading the example understand better what the field does).

examples/user_interaction/server/server.js
```diff
@@ -72,16 +72,25 @@ function handleRequest(request, response) {
...
     response.setHeader('Access-Control-Allow-Origin', '*');
-    response.setHeader('Access-Control-Allow-Methods', 'GET, POST, OPTIONS, PUT, PATCH, DELETE');
-    response.setHeader('Access-Control-Allow-Headers', 'X-Requested-With,content-type');
+    response.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
+    response.setHeader('Access-Control-Allow-Headers', 'traceparent');
     response.setHeader('Access-Control-Allow-Credentials', true);
 
     let result = '';
     let code = 200;
+    // Logic related to CORS pre-flight requests.
```
draffensperger: Maybe say "Accept all CORS pre-flight requests"? Would it make sense to move the `response.setHeader` calls into this block of code too?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -50,6 +57,10 @@ export class InteractionTracker {
     [index: string]: number;
   } = {};
 
+  // Map intended to keep track of current XHR objects
+  // associated to a span.
+  private readonly xhrTasks = new Map<XHRWithUrl, Span>();
```
draffensperger: Should this be called `xhrSpans`?

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -186,6 +198,51 @@ export class InteractionTracker {
     });
   }
 
+  private interceptXhrTasks(task: AsyncTask) {
+    if (isTrackedTask(task) && task.target instanceof XMLHttpRequest) {
```
draffensperger: Optional: you can reduce the nesting of this function by negating the `if` and making it return early, e.g.:

```
if (!isTrackedTask(task)) return;
if (!task.target instanceof XMLHttpRequest) return;

const xhr = ...
```

packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -186,6 +198,51 @@ export class InteractionTracker {
...
+      } else if (xhr.readyState === XMLHttpRequest.DONE) {
+        this.endXhrSpan(xhr);
+      }
+    }
+  }
+
+  private setTraceparentContextHeader(xhr: XHRWithUrl, rootSpan: RootSpan) {
+    const xhrUrl = xhr.__zone_symbol__xhrURL;
+    const childSpan = rootSpan.startChildSpan({
+      // `__zone_symbol__xhrURL` is set by the Zone monkey-path.
```
draffensperger: Should this comment go above where `xhrUrl` is set?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -339,11 +388,22 @@ describe('InteractionTracker', () => {
   }
 
   function doHTTPRequest() {
+    // Set the ocTraceOrigins value so the `traceparent` context header is set.
+    (window as WindowWithOcwGlobals).ocTraceHeaderHostRegex = /.*/;
```
draffensperger: Could you do a test case of an XHR to an origin that does not match the regex to demonstrate both cases of that `if`?

crdgonzalezca: Done.

examples/user_interaction/server/server.js
crdgonzalezca: Moved only the `setHeader`s related to the CORS request.

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -279,13 +279,49 @@ describe('InteractionTracker', () => {
...
+    });
+  });
+
+  it('should handle HTTP requets and set Trace Context Header', done => {
+    // Set the ocTraceHeaderHostRegex value so the `traceparent` context header is set.
+    (window as WindowWithOcwGlobals).ocTraceHeaderHostRegex = /.*/;
+    const setRequestHeaderSpy = spyOn(
+      XMLHttpRequest.prototype,
+      'setRequestHeader'
+    ).and.callThrough();
```
draffensperger: Do you check that this is called with the correct header?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -279,13 +279,49 @@ describe('InteractionTracker', () => {
...
+      expect(rootSpan.name).toBe('test interaction');
+      expect(rootSpan.attributes['EventType']).toBe('click');
+      expect(rootSpan.attributes['TargetElement']).toBe(BUTTON_TAG_NAME);
+      expect(rootSpan.ended).toBeTruthy();
+      expect(rootSpan.duration).toBeGreaterThanOrEqual(XHR_TIME);
+      expect(rootSpan.duration).toBeLessThanOrEqual(XHR_TIME + TIME_BUFFER);
+
+      expect(rootSpan.spans.length).toBe(1);
+      const childSpan = rootSpan.spans[0];
+      expect(setRequestHeaderSpy).not.toHaveBeenCalled();
```
draffensperger: Could you add a comment to this to highlight that the header is not set?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -399,8 +433,8 @@ describe('InteractionTracker', () => {
         xhr.dispatchEvent(event);
       }, XHR_TIME);
     });
-
-    xhr.open('GET', '/test');
+    if (!urlRequest) urlRequest = '/test';
```
draffensperger: You can do this via a default parameter, so `function doHTTPRequest(urlRequest: '/test')`

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -387,9 +423,7 @@ describe('InteractionTracker', () => {
     });
   }
 
-  function doHTTPRequest() {
-    // Set the ocTraceOrigins value so the `traceparent` context header is set.
-    (window as WindowWithOcwGlobals).ocTraceHeaderHostRegex = /.*/;
+  function doHTTPRequest(urlRequest?: string) {
```
draffensperger: Nit: could you call this `doHttpRequest`?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
crdgonzalezca: Yeah, this is done some lines below with `toHaveBeenCalledWith`

PR #132, opened 2019-07-15
Join browser performance data with the XHR spans

crdgonzalezca:
- Add `perf-resource-timing-selector` module to filter and get the `performance resource timings` related to the XHRs.
- Create a child span associated to the XHR span. This spans refers to the CORS pre-flight request in case the XHR caused this kind of request.
- Add performance data annotations to the XHR span.
- Move the XHR interceptor to a new separate module `xhr-interceptor` which allows a better control of intercepted XHRs and counting the amount of tasks being intercepted currently to clear the performance resource timing buffer.
- Update and add testing.
- Update the `adaptRootSpan` in the exporter agent to adapt all its descendants instead of adapting only its direct children.

This is how the traces look like:
image

### DIFF
--- a/packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
@@ -19,37 +19,12 @@ import { XhrPerformanceResourceTiming } from './zone-types';
 
 /**
  * Get Browser's performance resource timing data associated to a XHR.
- * For this case, some XHR might have two or one performance resource
- * timings as one of them is CORS pre-flight request and the second is related
- * to the actual HTTP request.
- * The algorithm to select performance resource timings related to that xhr is
- * is composed in general by three steps:
- *
- * 1. Filter the Performance Resource Timings by the name (it should match the
- * XHR URL), additionally, the start/end timings of every performance entry
- * should fit within the span start/end timings. These filtered performance
- * resource entries are considered as possible entries associated to the xhr.
- * Those are possible as there might be more than two entries that pass the
- * filter.
- *
- * 2. As the XHR could cause a CORS pre-flight, we have to look for either
- * possible pairs of performance resource timings or a single performance
- * resource entry (a possible pair is when a resource timing entry does not
- * overlap timings with other resource timing entry. Also, a possible single
- * resource timing is when that resource timing entry is not paired with any
- * other entry). Thus, for this step traverse the array of possible resource
- * entries and for every entry try to pair it with the other possible entries.
- *
- * 3. Pick the best performance resource timing for the XHR: Using the possible
- * performance resource timing entries from previous step, the best entry will
- * be the one with the minimum gap to the span start/end timings. That is the
- * substraction between the entry `respondeEnd` value and the span
- * `endPerfTime` plus the substraction between the entry `startTime` and span
- * `startPerfTime`. In case it is a tuple, the `startTime` corresponds to the
- * first entry and the `responseEnd` is from second entry.
- * The performance resource timing entry with the minimum gap to the span
- * start/end timings points out that entry is the best fit for the span.
- *
+ * Some XHRs might have two or one performance resource timings as one of them
+ * is the CORS pre-flight request and the second is related to the actual HTTP
+ * request.
+ * In overall, the algorithm to get this data takes the best fit for the span,
+ * this means the closest performance resource timing to the span start/end
+ * performance times is the returned value.
  * @param xhrUrl
  * @param span
  */
@@ -63,9 +38,17 @@ export function getXhrPerfomanceData(
   return bestEntry;
 }
 
-// Get Performance Resource Timings and filter them by matching the XHR url
-// with perfomance entry name. Additionally, the entry's start/end
-// timings must fit with in the span's start/end timings.
+/**
+ * First step for the algorithm. Filter the Performance Resource Timings by the
+ * name (it should match the XHR URL), additionally, the start/end timings of
+ * every performance entry should fit within the span start/end timings.
+ * These filtered performance resource entries are considered as possible
+ * entries associated to the xhr.
+ * Those are possible because there might be more than two entries that pass the
+ * filter.
+ * @param xhrUrl
+ * @param span
+ */
 export function getPerfResourceEntries(
   xhrUrl: string,
   span: Span
@@ -77,64 +60,86 @@ export function getPerfResourceEntries(
     ) as PerformanceResourceTiming[];
 }
 
+/**
+ * Second step for the 'Performance resource timings selector algorithm'.
+ * As the XHR could cause a CORS pre-flight request, we have to look for
+ * possible performance entries either containing cors preflight timings or not.
+ * A possible entry with cors data is when a resource timing entry does not
+ * overlap timings with other resource timing entry. Also, a possible entry
+ * without cors resource timing is when that resource timing entry is not
+ * 'paired' with any other entry.
+ * Thus, for this step traverse the array of resource entries and for every
+ * entry check if it is a possible performance resource entry.
+ * @param perfEntries
+ */
 export function getPossiblePerfResourceEntries(
   perfEntries: PerformanceResourceTiming[]
 ): XhrPerformanceResourceTiming[] {
   const possiblePerfEntries = new Array<XhrPerformanceResourceTiming>();
   const pairedEntries = new Set<PerformanceResourceTiming>();
-  let perfEntry1: PerformanceResourceTiming;
-  let perfEntry2: PerformanceResourceTiming;
+  let entryI: PerformanceResourceTiming;
+  let entryJ: PerformanceResourceTiming;
   // As this part of the algorithm traverses the array twice, although,
-  // this array is not big as the performance resource entries is cleared
-  // when there are no more running XHRs.
+  // this array is not large as the performance resource entries buffer is
+  // cleared when there are no more running XHRs.
   for (let i = 0; i < perfEntries.length; i++) {
-    perfEntry1 = perfEntries[i];
+    entryI = perfEntries[i];
     // Compare every performance entry with its consecutive perfomance entries.
     // That way to avoid comparing twice the entries.
     for (let j = i + 1; j < perfEntries.length; j++) {
-      perfEntry2 = perfEntries[j];
-      if (!overlappingPerfResourceTimings(perfEntry1, perfEntry2)) {
+      entryJ = perfEntries[j];
+      if (!overlappingPerfResourceTimings(entryI, entryJ)) {
         // As the entries are not overlapping, that means those timings
         // are possible perfomance timings related to the XHR.
-        possiblePerfEntries.push([perfEntry1, perfEntry2]);
-        pairedEntries.add(perfEntry1);
-        pairedEntries.add(perfEntry2);
+        possiblePerfEntries.push({
+          corsPreFlightRequest: entryI,
+          mainRequest: entryJ,
+        });
+        pairedEntries.add(entryI);
+        pairedEntries.add(entryJ);
       }
     }
-    // If the entry1 couldn't be paired with any other resource timing,
-    // add it as a single resource timing. This is possible because this
-    // single entry might be better that the other possible entries.
-    if (!pairedEntries.has(perfEntry1)) {
-      possiblePerfEntries.push(perfEntry1 as PerformanceResourceTiming);
+    // If the entry couldn't be paired with any other resource timing,
+    // add it as a possible resource timing without cors preflight data.
+    // This is possible because this entry might be better than the other
+    // possible entries.
+    if (!pairedEntries.has(entryI)) {
+      possiblePerfEntries.push({ mainRequest: entryI });
     }
   }
   return possiblePerfEntries;
 }
 
-// The best Performance Resource Timing Entry is considered the one with the
-// minimum gap the span end/start timings. That way we think that it fits
-// better to the XHR as it is the closest data to the actual XHR.
+// Pick the best performance resource timing for the XHR: Using the possible
+// performance resource timing entries from previous step, the best entry will
+// be the one with the minimum gap to the span start/end timings.
+// The performance resource timing entry with the minimum gap to the span
+// start/end timings points out that entry is the best fit for the span.
 function getBestPerfResourceTiming(
   perfEntries: XhrPerformanceResourceTiming[],
   span: Span
 ): XhrPerformanceResourceTiming | undefined {
   let minimumGapToSpan = Number.MAX_VALUE;
   let bestPerfEntry: XhrPerformanceResourceTiming | undefined = undefined;
-  let sumGapsToSpan: number;
   for (const perfEntry of perfEntries) {
-    // As a Tuple is in the end an Array, check that perfEntry is instance of
-    // Array is enough to know if this value refers to a Tuple.
-    if (perfEntry instanceof Array) {
-      sumGapsToSpan = Math.abs(perfEntry[0].startTime - span.startPerfTime);
-      sumGapsToSpan += Math.abs(perfEntry[1].responseEnd - span.endPerfTime);
+    let gapToSpan = Math.abs(
+      perfEntry.mainRequest.responseEnd - span.endPerfTime
+    );
+    // If the current entry has cors preflight data use its `startTime` to
+    // calculate the gap to the span.
+    if (perfEntry.corsPreFlightRequest) {
+      gapToSpan += Math.abs(
+        perfEntry.corsPreFlightRequest.startTime - span.startPerfTime
+      );
     } else {
-      sumGapsToSpan = Math.abs(perfEntry.responseEnd - span.endPerfTime);
-      sumGapsToSpan += Math.abs(perfEntry.startTime - span.startPerfTime);
+      gapToSpan += Math.abs(
+        perfEntry.mainRequest.startTime - span.startPerfTime
+      );
     }
     // If there is a new minimum gap to the span, update the minimum and pick
     // the current performance entry as the best at this point.
-    if (sumGapsToSpan < minimumGapToSpan) {
-      minimumGapToSpan = sumGapsToSpan;
+    if (gapToSpan < minimumGapToSpan) {
+      minimumGapToSpan = gapToSpan;
       bestPerfEntry = perfEntry;
     }
   }
--- a/packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
@@ -16,6 +16,7 @@
 
 import { Span } from '@opencensus/web-core';
 import { XhrPerformanceResourceTiming } from './zone-types';
+import { alreadyAssignedPerfEntries } from './xhr-interceptor';
 
 /**
  * Get Browser's performance resource timing data associated to a XHR.
@@ -32,8 +33,10 @@ export function getXhrPerfomanceData(
   xhrUrl: string,
   span: Span
 ): XhrPerformanceResourceTiming | undefined {
-  const filteredPerfEntries = getPerfResourceEntries(xhrUrl, span);
-  const possibleEntries = getPossiblePerfResourceEntries(filteredPerfEntries);
+  const filteredSortedPerfEntries = getPerfSortedResourceEntries(xhrUrl, span);
+  const possibleEntries = getPossiblePerfResourceEntries(
+    filteredSortedPerfEntries
+  );
   const bestEntry = getBestPerfResourceTiming(possibleEntries, span);
   return bestEntry;
 }
@@ -41,15 +44,18 @@ export function getXhrPerfomanceData(
 /**
  * First step for the algorithm. Filter the Performance Resource Timings by the
  * name (it should match the XHR URL), additionally, the start/end timings of
- * every performance entry should fit within the span start/end timings.
+ * every performance entry should fit within the span start/end timings. Also,
+ * the entry should be already assigned to a span.
  * These filtered performance resource entries are considered as possible
  * entries associated to the xhr.
  * Those are possible because there might be more than two entries that pass the
  * filter.
+ * Additionally, the returned array is sorted by the entries' `startTime` as
+ * getEntriesByType() already does it.
  * @param xhrUrl
  * @param span
  */
-export function getPerfResourceEntries(
+export function getPerfSortedResourceEntries(
   xhrUrl: string,
   span: Span
 ): PerformanceResourceTiming[] {
@@ -70,41 +76,35 @@ export function getPerfResourceEntries(
  * 'paired' with any other entry.
  * Thus, for this step traverse the array of resource entries and for every
  * entry check if it is a possible performance resource entry.
- * @param perfEntries
+ * @param filteredSortedPerfEntries Sorted array of Performance Resource
+ * Entries. This is sorted by the entries' `startTime`.
  */
 export function getPossiblePerfResourceEntries(
-  perfEntries: PerformanceResourceTiming[]
+  filteredSortedPerfEntries: PerformanceResourceTiming[]
 ): XhrPerformanceResourceTiming[] {
   const possiblePerfEntries = new Array<XhrPerformanceResourceTiming>();
-  const pairedEntries = new Set<PerformanceResourceTiming>();
-  let entryI: PerformanceResourceTiming;
-  let entryJ: PerformanceResourceTiming;
   // As this part of the algorithm traverses the array twice, although,
   // this array is not large as the performance resource entries buffer is
   // cleared when there are no more running XHRs.
-  for (let i = 0; i < perfEntries.length; i++) {
-    entryI = perfEntries[i];
-    // Compare every performance entry with its consecutive perfomance entries.
-    // That way to avoid comparing twice the entries.
-    for (let j = i + 1; j < perfEntries.length; j++) {
-      entryJ = perfEntries[j];
-      if (!overlappingPerfResourceTimings(entryI, entryJ)) {
+  for (let i = 0; i < filteredSortedPerfEntries.length; i++) {
+    const entryI = filteredSortedPerfEntries[i];
+    // Consider the current entry as a possible entry without cors preflight
+    // request. This is valid as this entry might be better than other possible
+    // entries.
+    possiblePerfEntries.push({ mainRequest: entryI });
+    // Compare every performance entry with the perfomance entries in front of
+    // it. This is possible as the entries are sorted by the startTime. That
+    // way we to avoid comparing twice the entries and taking the wrong order.
+    for (let j = i + 1; j < filteredSortedPerfEntries.length; j++) {
+      const entryJ = filteredSortedPerfEntries[j];
+      if (!isPossibleCorsPair(entryI, entryJ)) {
         // As the entries are not overlapping, that means those timings
         // are possible perfomance timings related to the XHR.
         possiblePerfEntries.push({
           corsPreFlightRequest: entryI,
           mainRequest: entryJ,
         });
-        pairedEntries.add(entryI);
-        pairedEntries.add(entryJ);
       }
-    }
-    // If the entry couldn't be paired with any other resource timing,
-    // add it as a possible resource timing without cors preflight data.
-    // This is possible because this entry might be better than the other
-    // possible entries.
-    if (!pairedEntries.has(entryI)) {
-      possiblePerfEntries.push({ mainRequest: entryI });
     }
   }
   return possiblePerfEntries;
@@ -122,20 +122,16 @@ function getBestPerfResourceTiming(
   let minimumGapToSpan = Number.MAX_VALUE;
   let bestPerfEntry: XhrPerformanceResourceTiming | undefined = undefined;
   for (const perfEntry of perfEntries) {
-    let gapToSpan = Math.abs(
-      perfEntry.mainRequest.responseEnd - span.endPerfTime
-    );
     // If the current entry has cors preflight data use its `startTime` to
     // calculate the gap to the span.
-    if (perfEntry.corsPreFlightRequest) {
-      gapToSpan += Math.abs(
-        perfEntry.corsPreFlightRequest.startTime - span.startPerfTime
-      );
-    } else {
-      gapToSpan += Math.abs(
-        perfEntry.mainRequest.startTime - span.startPerfTime
-      );
-    }
+    const perfEntryStartTime = perfEntry.corsPreFlightRequest
+      ? perfEntry.corsPreFlightRequest.startTime
+      : perfEntry.mainRequest.startTime;
+    const gapToSpan =
+      span.endPerfTime -
+      perfEntry.mainRequest.responseEnd +
+      (perfEntryStartTime - span.startPerfTime);
+
     // If there is a new minimum gap to the span, update the minimum and pick
     // the current performance entry as the best at this point.
     if (gapToSpan < minimumGapToSpan) {
@@ -152,13 +148,14 @@ function isPerfEntryPartOfXhr(
   span: Span
 ): boolean {
   return (
+    !alreadyAssignedPerfEntries.has(entry) &&
     entry.name === xhrUrl &&
     entry.startTime >= span.startPerfTime &&
     entry.responseEnd <= span.endPerfTime
   );
 }
 
-function overlappingPerfResourceTimings(
+function isPossibleCorsPair(
   entry1: PerformanceResourceTiming,
   entry2: PerformanceResourceTiming
 ): boolean {
--- a/packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
@@ -1,0 +1,164 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Span } from '@opencensus/web-core';
+import { XhrPerformanceResourceTiming } from './zone-types';
+
+/**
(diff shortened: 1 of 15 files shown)


packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,164 @@
...
+      isPerfEntryPartOfXhr(entry as PerformanceResourceTiming, xhrUrl, span)
+    ) as PerformanceResourceTiming[];
+}
+
+export function getPossiblePerfResourceEntries(
+  perfEntries: PerformanceResourceTiming[]
+): XhrPerformanceResourceTiming[] {
+  const possiblePerfEntries = new Array<XhrPerformanceResourceTiming>();
+  const pairedEntries = new Set<PerformanceResourceTiming>();
+  let perfEntry1: PerformanceResourceTiming;
```
draffensperger: Would it work to do a `for (const .. of` loop through the `perfEntries` to populate this variable and also `perfEntry2`?

Also, what would you think about calling the variables something like `entryI` and `entryJ`?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,164 @@
...
+  let perfEntry2: PerformanceResourceTiming;
+  // As this part of the algorithm traverses the array twice, although,
```
draffensperger: Would it be possible to use sorting to find the non-overlapping intervals more efficiently? See http://web.stanford.edu/class/archive/cs/cs161/cs161.1176/Sections/final_review-1.pdf

In practice this algorithm would be fine since the data size is small, but worth looking at that sorting-based one. It may even make the code simpler and shorter!

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -67,3 +67,24 @@ export declare interface WindowWithOcwGlobals extends Window {
...
   ocTraceHeaderHostRegex?: string | RegExp;
 }
+
+/**
+ * Allows to keep track of performance entries related to a XHR.
+ * As some XHRs might generate a CORS pre-flight request, the XHR
+ * might have either two performance resource entries or a single
+ * performance resource entry.
+ */
+export type XhrPerformanceResourceTiming =
```
draffensperger: Rather than making this either a tuple or timing, what would you think about making this always an array?

My thinking is that if it's always an array then some of the types and code could be easier to understand. It also unifies the way you compare start/end time in the pair case and the single request case (both cases use the start time of the first element and the end time of the last element)

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,164 @@
...
+): XhrPerformanceResourceTiming | undefined {
+  let minimumGapToSpan = Number.MAX_VALUE;
+  let bestPerfEntry: XhrPerformanceResourceTiming | undefined = undefined;
+  let sumGapsToSpan: number;
+  for (const perfEntry of perfEntries) {
+    // As a Tuple is in the end an Array, check that perfEntry is instance of
+    // Array is enough to know if this value refers to a Tuple.
+    if (perfEntry instanceof Array) {
+      sumGapsToSpan = Math.abs(perfEntry[0].startTime - span.startPerfTime);
+      sumGapsToSpan += Math.abs(perfEntry[1].responseEnd - span.endPerfTime);
```
draffensperger: Could you do something like `perfEntry[perfEntry.length - 1]` here and remove the `} else {` block by making it so that the paired case (with CORS preflight) and non-pair cases both use a list?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,164 @@
...
+  span: Span
+): XhrPerformanceResourceTiming | undefined {
+  let minimumGapToSpan = Number.MAX_VALUE;
+  let bestPerfEntry: XhrPerformanceResourceTiming | undefined = undefined;
+  let sumGapsToSpan: number;
+  for (const perfEntry of perfEntries) {
+    // As a Tuple is in the end an Array, check that perfEntry is instance of
+    // Array is enough to know if this value refers to a Tuple.
+    if (perfEntry instanceof Array) {
+      sumGapsToSpan = Math.abs(perfEntry[0].startTime - span.startPerfTime);
```
draffensperger: What would you think about changing `sumGapsToSpan` to `const` and making it the result of the two values added up?

That could be done via a helper function or via the refactor to make both cases (with and without preflight) use an array type. In general, I prefer to vaoid `let` unless needed for efficiently in an algorithm.

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,151 @@
...
+import {
+  annotationsForPerfTimeFields,
+  PerformanceResourceTimingExtended,
+  PERFORMANCE_ENTRY_EVENTS,
+  getResourceSpan,
+} from '@opencensus/web-instrumentation-perf';
+
+// Map intended to keep track of current XHR objects
+// associated to a span.
+const xhrSpans = new Map<XHRWithUrl, Span>();
```
draffensperger: Nit: should this be named as `XhrWithUrl` to match the Google casing conventions?

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,151 @@
...
+  const childSpan = rootSpan.startChildSpan({
+    name: parseUrl(xhrUrl).pathname,
+    kind: SpanKind.CLIENT,
+  });
+  // Associate the child span to the XHR so it allows to
+  // find the correct span when the request is DONE.
+  xhrSpans.set(xhr, childSpan);
+  if (traceOriginMatchesOrSameOrigin(xhrUrl)) {
+    xhr.setRequestHeader(
+      'traceparent',
```
draffensperger: Could you make this a file level constant, e.g. `const TRACEPARENT_HEADER = 'traceparent';`? I say that because I think OpenCensus Node/Open Telemetry JS do not, mostly as a stylistic thing I think since the header itself is not likely to change!

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,151 @@
...
+
+function joinPerfResourceDataToSpan(xhr: XHRWithUrl, span: Span) {
+  const perfResourceTimings = getXhrPerfomanceData(xhr.responseURL, span);
+  if (perfResourceTimings instanceof Array) {
+    // This case is true when the resource timings data associates two entries
+    // to the span, where the first entry is the CORS pre-flight request and
+    // the second is the actual HTTP request. Create a child span which is
+    // related to the CORS pre-flight and use the second entry to add
+    // annotations to the span.
+    const corsPerfTiming = perfResourceTimings[0] as PerformanceResourceTimingExtended;
```
draffensperger: Would it be worth while to use destructuring here? E.g.:
```
const [corsPerfTiming, actualXhrPerfTiming] = perfResourceTimings as PerformanceResourceTimingExtended[];
```

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,151 @@
...
+    );
+  }
+}
+
+function setCorsPerfTimingAsChildSpan(
+  performanceTiming: PerformanceResourceTimingExtended,
+  span: Span
+): void {
+  const corsSpan = getResourceSpan(performanceTiming, span.traceId, span.id);
+  corsSpan.name = 'CORS';
```
draffensperger: What would you think about calling this `'CORS Preflight'`?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -287,8 +294,9 @@ describe('InteractionTracker', () => {
       XMLHttpRequest.prototype,
       'setRequestHeader'
     ).and.callThrough();
+    const urlRequest = 'http://localhost:8000/test';
```
draffensperger: Nit: should this be `requestUrl`?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -308,6 +316,22 @@ describe('InteractionTracker', () => {
       expect(childSpan.attributes[ATTRIBUTE_HTTP_STATUS_CODE]).toBe('200');
       expect(childSpan.attributes[ATTRIBUTE_HTTP_METHOD]).toBe('GET');
       expect(childSpan.ended).toBeTruthy();
+      const childSpanperformanceEntries = getXhrPerfomanceData(
```
draffensperger: Nit: capitalization of `performance` here so, `childSpanPerformanceEntries`?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -308,6 +316,22 @@ describe('InteractionTracker', () => {
...
+        urlRequest,
+        childSpan
+      );
+      expect(childSpanperformanceEntries).toBeTruthy();
+      if (childSpanperformanceEntries) {
+        expect(childSpan.annotations).toEqual(
+          annotationsForPerfTimeFields(
```
draffensperger: What would you think about just listing out the hard-coded annotations here? That way we are checking the values directly rather than relying on the correctness of `annotationsForPerfTimeFields`, which in principle could have a bug in both the test and the actual code

This applies to below cases as well

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -392,6 +437,25 @@ describe('InteractionTracker', () => {
       expect(childSpan.attributes[ATTRIBUTE_HTTP_STATUS_CODE]).toBe('200');
       expect(childSpan.attributes[ATTRIBUTE_HTTP_METHOD]).toBe('GET');
       expect(childSpan.ended).toBeTruthy();
+      const childSpanPerfEntries = getXhrPerfomanceData(urlRequest, childSpan);
+      expect(childSpanPerfEntries).toBeTruthy();
+      if (childSpanPerfEntries instanceof Array) {
```
draffensperger: Rather than checking here, can we know what it should be and `expect` it? (I think that applies to a case above too)

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,216 @@
...
+import {
+  getPerfResourceEntries,
+  getPossiblePerfResourceEntries,
+  getXhrPerfomanceData,
+} from '../src/perf-resource-timing-selector';
+import { spyPerfEntryByType, createFakePerfResourceEntry } from './util';
+
+describe('Perf Resource Timing Selector', () => {
+  describe('getPerfResourceEntries', () => {
+    it('Should filter properly the resource entries', () => {
```
draffensperger: Could you specify a bit more what this is filtering for?

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,216 @@
...
+
+      const entriesToBeFiltered = [entry2, entry3];
+      const filteredPerfEntries = getPerfResourceEntries('/test', span);
+      expect(filteredPerfEntries).toEqual(entriesToBeFiltered);
+    });
+  });
+
+  describe('getPossiblePerfResourceEntries', () => {
+    it('Should pair no overlapping entries', () => {
+      const entry1 = createFakePerfResourceEntry(
```
draffensperger: What would you think about calling these something like `overlappingEntry1` and `nonOverlappingEntry2`, etc. to indicate a bit more what the fake data is showing?

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,216 @@
...
+      // endTime.
+      expect(filteredPerfEntries).not.toContain([entry1, entry2]);
+      expect(filteredPerfEntries).toContain([entry1, entry3]);
+      expect(filteredPerfEntries).toContain([entry1, entry4]);
+      expect(filteredPerfEntries).not.toContain([entry2, entry3]);
+      expect(filteredPerfEntries).toContain([entry2, entry4]);
+      expect(filteredPerfEntries).not.toContain([entry3, entry4]);
+    });
+
+    it('Should take single resource timing entries', () => {
```
draffensperger: Could you clarify this a bit? What does it mean to take a single timing entry?

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,216 @@
...
+      const filteredPerfEntries = getPossiblePerfResourceEntries(
+        perfResourceEntries
+      );
+      expect(filteredPerfEntries.length).toBe(3);
+      expect(filteredPerfEntries).toContain(entry1);
+      expect(filteredPerfEntries).toContain(entry2);
+      expect(filteredPerfEntries).toContain(entry3);
+    });
+
+    it('Should take single and tuples of resource timing entries', () => {
```
draffensperger: How would this test change if we make both cases use arrays? Is there still a useful test case here?

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,216 @@
...
+      expect(filteredPerfEntries).not.toContain(entry3);
+      // As entry4 has not been paired, should contain this a single entry.
+      expect(filteredPerfEntries).toContain(entry4);
+    });
+  });
+
+  describe('getXhrPerfomanceData', () => {
+    it('Should take the tuple with the minimum gap to the span as the best resource timing entry', () => {
+      const entry1 = createFakePerfResourceEntry(
+        13.100000011036173,
```
draffensperger: Optional/nit: what would you think about simplifying the numbers a bit here to make the gap differences more obvious?

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,216 @@
...
+      );
+      const perfResourceEntries = [entry1, entry2, entry3, entry4, entry5];
+      spyPerfEntryByType(perfResourceEntries);
+      const span = createSpan(13.1, 18.3);
+
+      const filteredPerfEntries = getXhrPerfomanceData('/test', span);
+      expect(filteredPerfEntries).toEqual([entry1, entry4]);
+    });
+
+    it('Should take the single resource entry with the minimum gap to the span as the best entry', () => {
```
draffensperger: How is this case different from the one above?

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
johnbryan: Another option might be an object with fields like
{
  preFlightRequest?: PerformanceResourceTiming,  // optional
  mainRequest: PerformanceResourceTiming,
}

That might make it more explicit to the reader what each piece means, vs using an array index. Each approach will have its pros and cons so up to you what seems best!

draffensperger: That's a good idea too - I didn't think of an object based approach!

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
crdgonzalezca: Well it would work but it would add more complexity to check the possible performance entry as using this case, what we would now check is that the entries have not been compared before, as we would be doing the process twice. I would like to keep the loops in this way as I can control the entries you are comparing.

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
crdgonzalezca: I think this does not work, as what we actually want is to check every possible pair and see if they don't overlap to count it as possible xhr performance entry. However, the idea of the suggested algorithm is a way to make a non overlapping array but we don't actually want to make the entries not overlapping as this might return the wrong entry associated to the XHR.

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
crdgonzalezca: I decided for the object idea, as I think makes the code easier to read and better to explain the algorithm steps and tests.

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
crdgonzalezca: As part of taking the XhrPerformaceResourceTiming type as an object, instead of labeling these objects like that, I decided to create the expected (non)overlapping entries following.

packages/opencensus-web-instrumentation-zone/test/test-perf-resource-timing-selector.ts
crdgonzalezca: This test case is different in the sense that the returned value does not have a cors preflight request value as this is another possible return value. I made this more clear.

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+    entryI = perfEntries[i];
+    // Compare every performance entry with its consecutive perfomance entries.
+    // That way to avoid comparing twice the entries.
+    for (let j = i + 1; j < perfEntries.length; j++) {
+      entryJ = perfEntries[j];
+      if (!overlappingPerfResourceTimings(entryI, entryJ)) {
+        // As the entries are not overlapping, that means those timings
+        // are possible perfomance timings related to the XHR.
+        possiblePerfEntries.push({
+          corsPreFlightRequest: entryI,
```
johnbryan: this assumes that perfEntries is sorted (which is valid because performance.getEntriesByType() returns sorted data). But it would be good to clarify this in a comment somewhere, maybe in the @param line? Or perhaps call filteredPerfEntries "filteredSortedPerfEntries"?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+// start/end timings points out that entry is the best fit for the span.
+function getBestPerfResourceTiming(
+  perfEntries: XhrPerformanceResourceTiming[],
+  span: Span
+): XhrPerformanceResourceTiming | undefined {
+  let minimumGapToSpan = Number.MAX_VALUE;
+  let bestPerfEntry: XhrPerformanceResourceTiming | undefined = undefined;
+  for (const perfEntry of perfEntries) {
+    let gapToSpan = Math.abs(
+      perfEntry.mainRequest.responseEnd - span.endPerfTime
```
johnbryan: we already checked earlier that responseEnd is < span.endPerfTime, right? So could the "Math.abs" here and below be removed and just do the subtraction in the proper order?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+    );
+    // If the current entry has cors preflight data use its `startTime` to
+    // calculate the gap to the span.
+    if (perfEntry.corsPreFlightRequest) {
```
johnbryan: This is ok as a conditional, but could be a bit shorter with something like:

const perfStart = perfEntry.corsPreFlightRequest ? perfEntry.corsPreFlightRequest.startTime : perfEntry.mainRequest.startTime;
gapToSpan += perfStart - span.startPerfTime

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+  perfEntries: PerformanceResourceTiming[]
+): XhrPerformanceResourceTiming[] {
+  const possiblePerfEntries = new Array<XhrPerformanceResourceTiming>();
+  const pairedEntries = new Set<PerformanceResourceTiming>();
+  let entryI: PerformanceResourceTiming;
+  let entryJ: PerformanceResourceTiming;
+  // As this part of the algorithm traverses the array twice, although,
+  // this array is not large as the performance resource entries buffer is
+  // cleared when there are no more running XHRs.
+  for (let i = 0; i < perfEntries.length; i++) {
```
draffensperger: Can you use `for (const .. of` here?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+    entryI = perfEntries[i];
+    // Compare every performance entry with its consecutive perfomance entries.
+    // That way to avoid comparing twice the entries.
+    for (let j = i + 1; j < perfEntries.length; j++) {
+      entryJ = perfEntries[j];
```
draffensperger: Can you move the `const` here?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+ * 'paired' with any other entry.
+ * Thus, for this step traverse the array of resource entries and for every
+ * entry check if it is a possible performance resource entry.
+ * @param perfEntries
+ */
+export function getPossiblePerfResourceEntries(
+  perfEntries: PerformanceResourceTiming[]
+): XhrPerformanceResourceTiming[] {
+  const possiblePerfEntries = new Array<XhrPerformanceResourceTiming>();
+  const pairedEntries = new Set<PerformanceResourceTiming>();
```
draffensperger: Could you add a comment here on what this tracks? Should it maybe be called "corsPairs"?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+        });
+        pairedEntries.add(entryI);
+        pairedEntries.add(entryJ);
+      }
+    }
+    // If the entry couldn't be paired with any other resource timing,
+    // add it as a possible resource timing without cors preflight data.
+    // This is possible because this entry might be better than the other
+    // possible entries.
+    if (!pairedEntries.has(entryI)) {
```
draffensperger: Can we remove this `if` statement and still have the algorithm work? I'm not 100% sure yet that it is always correct to have the if, and anyway removing it could make it simpler - I'm open if I'm wrong on this!

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+  // As this part of the algorithm traverses the array twice, although,
+  // this array is not large as the performance resource entries buffer is
+  // cleared when there are no more running XHRs.
+  for (let i = 0; i < perfEntries.length; i++) {
+    entryI = perfEntries[i];
+    // Compare every performance entry with its consecutive perfomance entries.
+    // That way to avoid comparing twice the entries.
+    for (let j = i + 1; j < perfEntries.length; j++) {
+      entryJ = perfEntries[j];
+      if (!overlappingPerfResourceTimings(entryI, entryJ)) {
```
draffensperger: Could you also explain that the performance timing entries are implicitly sorted by end time?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+  span: Span
+): boolean {
+  return (
+    entry.name === xhrUrl &&
+    entry.startTime >= span.startPerfTime &&
+    entry.responseEnd <= span.endPerfTime
+  );
+}
+
+function overlappingPerfResourceTimings(
```
draffensperger: Would it make sense to call this `isPossibleCorsPair`?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,169 @@
...
+ * Some XHRs might have two or one performance resource timings as one of them
+ * is the CORS pre-flight request and the second is related to the actual HTTP
+ * request.
+ * In overall, the algorithm to get this data takes the best fit for the span,
+ * this means the closest performance resource timing to the span start/end
+ * performance times is the returned value.
+ * @param xhrUrl
+ * @param span
+ */
+export function getXhrPerfomanceData(
```
draffensperger: Would it make sense to keep a `Set` of already assigned performance entries? If so, it should be cleared when you clear the performance entry buffer. Then you could skip out on anything already assigned

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
crdgonzalezca: My decision was to remove this part and do not use this Set anymore, this to remove complexity on the algorithm. However it still is working as expected.

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
crdgonzalezca: It is not possible as we need to look the possible pair for the entryI comparing with the entries in front of it. This is possible as the entries array is already sorted with the `startTime`. Added some comments.

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * First step for the algorithm. Filter the Performance Resource Timings by the
+ * name (it should match the XHR URL), additionally, the start/end timings of
+ * every performance entry should fit within the span start/end timings. Also,
+ * the entry should be already assigned to a span.
+ * These filtered performance resource entries are considered as possible
+ * entries associated to the xhr.
+ * Those are possible because there might be more than two entries that pass the
+ * filter.
+ * Additionally, the returned array is sorted by the entries' `startTime` as
```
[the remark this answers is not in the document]
crdgonzalezca: Yes, here is the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType#Return_Value) on the returned value by `getEntriesByType`.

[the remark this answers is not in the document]
draffensperger: Cool. Would it make sense to link to that documentation?

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,157 @@
...
+  corsSpan.name = 'CORS Preflight';
+  span.spans.push(corsSpan);
+}
+
+function incrementXhrTaskCount(): void {
+  xhrTasksCount++;
+}
+
+function decrementXhrTaskCount(): void {
+  if (xhrTasksCount > 0) xhrTasksCount--;
```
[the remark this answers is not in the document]
crdgonzalezca: Not actually, my though was to just make sure we are not decreasing to negative numbers. This case should never happen. I think this makes clear that is not possible to decrease below 0.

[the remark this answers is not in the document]
draffensperger: OK, makes sense. Could you just add a comment to explain that this is not expected to go below zero, but this check makes that clear?

examples/user_interaction/client/src/App.js
```diff
@@ -29,12 +29,14 @@ class App extends React.Component {
   handleClick() {
     // Use promises to test behavior on MicroTasks.
     const promise = new Promise(resolve => {
+      console.log(window.Zone.current);
```
draffensperger: Are these log statements meant to be temporary?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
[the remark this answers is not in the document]
crdgonzalezca: Yeah, added.

examples/user_interaction/client/src/App.js
crdgonzalezca: Sorry, I was doing something different. Not related, just deleted this.

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,183 @@
...
+ * name (it should match the XHR URL), additionally, the start/end timings of
+ * every performance entry should fit within the span start/end timings. Also,
+ * the entry should not be already assigned to a span.
+ * These filtered performance resource entries are considered as possible
+ * entries associated to the xhr.
+ * Those are possible because there might be more than two entries that pass the
+ * filter.
+ * Additionally, the returned array is sorted by the entries' `startTime` as
+ * getEntriesByType() already does it (https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntriesByType#Return_Value).
+ * @param xhrUrl
```
draffensperger: If there is no documentation for these params, could you remove them from the JSDoc?

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,183 @@
...
+}
+
+/**
+ * Pick the best performance resource timing for the XHR: Using the possible
+ * performance resource timing entries from previous step, the best entry will
+ * be the one with the minimum gap to the span start/end timings.
+ * The performance resource timing entry with the minimum gap to the span
+ * start/end timings points out that entry is the best fit for the span.
+ *
+ * @param perfEntries
```
draffensperger: Same here on removing these extra @param tags with no description

packages/opencensus-web-instrumentation-zone/src/perf-resource-timing-selector.ts
```diff
@@ -0,0 +1,183 @@
...
+
+/**
+ * A possible CORS pair is defined when the entries does not overlap in their
+ * start/end times.
+ */
+function isPossibleCorsPair(
+  entry1: PerformanceResourceTiming,
+  entry2: PerformanceResourceTiming
+): boolean {
+  // To determine if the entries overlap, the minimum responseEnd should be
```
draffensperger: Could you simplify this logic if you made `entry1` be called `maybePreflight` and `entry2` be called `maybeMainRequset`? The assumption there is that the `maybePreflight` start time is definitely less than the `maybeMainRequest` start time, because of the sorting done by `getEntriesByType`

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,158 @@
...
+import {
+  annotationsForPerfTimeFields,
+  PerformanceResourceTimingExtended,
+  PERFORMANCE_ENTRY_EVENTS,
+  getResourceSpan,
+} from '@opencensus/web-instrumentation-perf';
+
+const TRACEPARENT_HEADER = 'traceparent';
+
+// Map intended to keep track of current XHR objects
```
draffensperger: Nit: could you use JSDoc style commetns for these /** */ ?

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,158 @@
...
+// some already assigned entries.
+export const alreadyAssignedPerfEntries = new Set<PerformanceResourceTiming>();
+
+/**
+ * Intercepts task as XHR if it is a tracked task and its target object is
+ * instance of `XMLHttpRequest`.
+ * In case the task is intercepted, sets the Trace Context Header to it and
+ * creates a child span related to this XHR in case it is OPENED.
+ * In case the XHR is DONE, end the child span.
+ * @param task
```
draffensperger: Can this empty @param tag be removed?

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,158 @@
...
+    // than `string`. Once it is able to send as a number, change it.
+    span.addAttribute(ATTRIBUTE_HTTP_STATUS_CODE, xhr.status.toString());
+    span.addAttribute(ATTRIBUTE_HTTP_METHOD, xhr._ocweb_method);
+    span.end();
+    joinPerfResourceDataToSpan(xhr, span);
+    xhrSpans.delete(xhr);
+  }
+}
+
+// If xhr task count is 0, clear the Performance Resource Timings.
```
draffensperger: Nit: can this be JSDoc style /** */ ?

packages/opencensus-web-instrumentation-zone/src/xhr-interceptor.ts
```diff
@@ -0,0 +1,162 @@
...
+
+function joinPerfResourceDataToSpan(xhr: XhrWithUrl, span: Span) {
+  const xhrPerfResourceTiming = getXhrPerfomanceData(xhr.responseURL, span);
+  if (!xhrPerfResourceTiming) return;
+
+  alreadyAssignedPerfEntries.add(xhrPerfResourceTiming.mainRequest);
+  if (xhrPerfResourceTiming.corsPreFlightRequest) {
+    alreadyAssignedPerfEntries.add(xhrPerfResourceTiming.corsPreFlightRequest);
+    const corsPerfTiming = xhrPerfResourceTiming.corsPreFlightRequest as PerformanceResourceTimingExtended;
+    setCorsPerfTimingAsChildSpan(corsPerfTiming, span);
```
johnbryan: When there's a cors pre-flight request, would it make sense to add the main request as a child span too? (I'm not saying you definitely should, just curious what you think)

crdgonzalezca: I think that might be another option too, however, as the span is already related to the XHR I think we just want to see some performance annotation related to the main request and also, as we are sending the `traceparent` context header to the server, we will be able to see the span related to that specific request. otherwise, if the `traceparent` is not sent, the annotations will be enough I think.

PR #141, opened 2019-07-17
Custom spans

crdgonzalezca:
Update the user interaction client example creating custom spans for the different tasks the click handler runs.
Also, update a little the documentation on how to do it. I know the documentation is not complete now, however, it will be completely updated later in the next weeks for productionization, for now just update a bit the documentation on this.

### DIFF
--- a/examples/user_interaction/client/src/App.js
+++ b/examples/user_interaction/client/src/App.js
@@ -50,7 +50,9 @@ class App extends React.Component {
 
   callSleepApi() {
     const xhr = new XMLHttpRequest();
-    // Create a child span for the XHR.
+    // Create a child span for the XHR. It is possible to create your own spans 
+    // even if the involved task or operation already generates an automatic 
+    // span. In this case, automatic spans are generated for XHRs.
     const callSleepApiCustomSpan = tracing.tracer.startChildSpan({ name: 'Call Sleep API' });
     xhr.onreadystatechange = () => {
       if (xhr.readyState === XMLHttpRequest.DONE) {
--- a/examples/user_interaction/client/src/App.js
+++ b/examples/user_interaction/client/src/App.js
@@ -28,20 +28,22 @@ class App extends React.Component {
   }
 
   handleClick() {
-    // Start a child span for the Promise, this span will be son of the 
-    // current root span related to the current user interaction.
-    // These spans should be created in the code the click handler will run.
-    const childSpan = tracing.tracer.startChildSpan({ name: 'promise' });
     // Use promises to test behavior on MicroTasks.
     const promise = new Promise(resolve => {
+      // Start a child span for the setTimeout as this is the operation we want 
+      // to measure. 
+      // This span will be child of the  current root span related to the 
+      // current user interaction. Additionally, these spans should be created
+      // in the code the click handler will run.
+      const setTimeoutCustomSpan = tracing.tracer.startChildSpan({ name: 'setTimeout custom span' });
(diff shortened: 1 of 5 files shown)


examples/user_interaction/client/src/App.js
```diff
@@ -27,6 +28,10 @@ class App extends React.Component {
   }
 
   handleClick() {
+    // Start a child span for the Promise, this span will be son of the
```
draffensperger: Nit: could we say "child" rather than "son"?

examples/user_interaction/client/src/App.js
```diff
@@ -27,6 +28,10 @@ class App extends React.Component {
   }
 
   handleClick() {
+    // Start a child span for the Promise, this span will be son of the 
+    // current root span related to the current user interaction.
+    // These spans should be created in the code the click handler will run.
+    const childSpan = tracing.tracer.startChildSpan({ name: 'promise' });
```
draffensperger: Could we give a more descriptive name to this custom span? What is the operation doing that it is measuring? I would think the more descriptive name could be given to both the `childSpan` variable and the `name:` field

crdgonzalezca:
Requested changes are done. Ready to review.

examples/user_interaction/client/src/App.js
```diff
@@ -29,21 +30,32 @@ class App extends React.Component {
...
     promise.then(() => {
-      console.log("Resolving promise");
       this.callSleepApi();
     });
   }
 
   callSleepApi() {
     const xhr = new XMLHttpRequest();
+    // Create a child span for the XHR.
+    const callSleepApiCustomSpan = tracing.tracer.startChildSpan({ name: 'Call Sleep API' });
```
draffensperger: What would you think about removing this custom span since there will already be one for the XHR?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -280,6 +280,41 @@ describe('InteractionTracker', () => {
...
 
+  describe('Custom Spans', () => {
+    it('Should handle the custom spans and add them to the current root span as child spans', done => {
+      const onclick = () => {
+        // Start a custom span for the setTimeout.
+        const setTimeoutCustomSpan = tracing.tracer.startChildSpan({
+          name: 'setTimeout custom span',
+        });
+        setTimeout(() => {
+          noop();
```
draffensperger: Is this call needed?

examples/user_interaction/client/src/App.js
crdgonzalezca: I would like to keep it. My thought is to show in the example that is possible to create custom spans even if those tasks already generate automatic spans. I will add some comments to clarify more this example.

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
crdgonzalezca: Not actually, removed.

examples/user_interaction/client/src/App.js
draffensperger: OK, I think it's fine to keep. Is there another call that does not have a custom span on it?

crdgonzalezca: Well, the call to prime numbers API does not have a custom span, as it also is an XHR, I decided to not create a custom span for that.

PR #142, opened 2019-07-19
Monkey-Patch `XMLHttpRequest.send()` to send traceparent header and generate XHR span when the this method is called

crdgonzalezca:
I found out a bug in the xhr-interceptor when the XHR span is generated and the browser performance data is joined to the XHR span. The XHR span was being started when the XHR is in OPENED state (`open()` method is called), however, the http requests that the XHR generates are only done when the `send()` method is called. The problem here is that the method `send()` could be called a long time after  `open()` is called. This means that there might be a big gap between `open()` and `send()` calls, and within this gap, there might be requests from other interactions calling to the same URL, this will generate a wrong result in the performance resource timing selector, as those entries could have a better fit in the span.

The solution then is to monkey-patch `XMLHttpRequest.send()` to send the traceparent header and generate the XHR span when the this method is called, but before the HTTP requests are done.
This means that the Performance entries should fit almost exactly in the span's start/end times and the XHR span will actually measure the time taken by the XHR.

### DIFF
--- a/packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
@@ -26,7 +26,7 @@ import {
   RESET_TRACING_ZONE_DELAY,
 } from '../src/interaction-tracker';
 import { doPatching } from '../src/monkey-patching';
-import { WindowWithOcwGlobals } from '../src/zone-types';
+import { WindowWithOcwGlobals, XhrWithOcWebData } from '../src/zone-types';
 import { spanContextToTraceParent } from '@opencensus/web-propagation-tracecontext';
 import { createFakePerfResourceEntry, spyPerfEntryByType } from './util';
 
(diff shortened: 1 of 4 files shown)


packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -288,7 +288,7 @@ describe('InteractionTracker', () => {
       perfResourceEntries = [];
     });
 
-    it('should handle HTTP requets and do not set Trace Context Header', done => {
+    fit('should handle HTTP requets and do not set Trace Context Header', done => {
```
draffensperger: Could you change this back to `it`?

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -494,8 +494,19 @@ describe('InteractionTracker', () => {
       const xhr = new XMLHttpRequest();
       xhr.onreadystatechange = noop;
       spyOn(xhr, 'send').and.callFake(() => {
+        // Simulate the actual monkey-patch done on `Send()`.
```
draffensperger: Is there a way to avoid having to duplicate the monkey-patch logic here?

For example, can you do something like `const realSend = xhr.send` and then `xhr.send = () => { realSend(); ....}`?

crdgonzalezca: Isn't that mean that the actual HTTP requests will be sent? Isn't it something we want to avoid to call the actual `send`?

crdgonzalezca: Another option would be to export a function that runs the logic on the monkey-patch and called from the test.

draffensperger: Chatted offline - I like the idea of an exported function.

PR #143, opened 2019-07-22
Use CSS selector to name the interaction

crdgonzalezca:
Use CSS selector using the tag name and the ID to name the User interaction (e.g. button#id).

### DIFF
--- a/packages/opencensus-web-instrumentation-zone/src/history-api-patch.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/history-api-patch.ts
@@ -15,14 +15,7 @@
  */
 
 import { tracing } from '@opencensus/web-core';
-import { getTraceId } from './util';
-
-/**
- * Set of strings to store the trace Id of interactions that might change the
- * name during interaction tracking. This is the case when interaction name is
- * a CSS selector.
- */
-export const interactionsMightChangeName = new Set<string>();
+import { isRootSpanNameReplaceable } from './util';
 
 /**
  * Monkey-patch `History API` to detect route transitions. This is necessary
@@ -82,13 +75,11 @@ function maybeUpdateInteractionName(previousLocationPathname: string) {
   // that means the name might change to `Navigation <pathname>` as this is a
   // more understadable name for the interaction in case the location
   // pathname actually changed.
-  const traceId = getTraceId(Zone.current);
   if (
     rootSpan &&
-    interactionsMightChangeName.has(traceId) &&
+    isRootSpanNameReplaceable(Zone.current) &&
     previousLocationPathname !== location.pathname
   ) {
     rootSpan.name = 'Navigation ' + location.pathname;
-    interactionsMightChangeName.delete(traceId);
   }
 }
--- a/packages/opencensus-web-instrumentation-zone/src/history-api-patch.ts
+++ b/packages/opencensus-web-instrumentation-zone/src/history-api-patch.ts
@@ -1,0 +1,94 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


packages/opencensus-web-instrumentation-zone/src/interaction-tracker.ts
```diff
@@ -319,15 +319,14 @@ export class InteractionTracker {
...
-    // Also, we check if the location pathname did change.
+    // If for this interaction, the developer did not give any explicit
+    // attibute (`data-ocweb-id`) and the generated name is a CSS selector,
+    // that means the name might change to `Navigation <pathname>` as this is a
+    // more understadable name for the interaction in case the location
+    // pathname actually changed.
     if (
       rootSpan &&
-      rootSpan.name.startsWith('<') &&
+      document.querySelector(rootSpan.name) &&
```
draffensperger: What's the role of this check? What would happen if it were just removed?

crdgonzalezca: As the `data-ocweb-id` attribute has the highest priority to name the interaction, and the `Navigation` is in second priority, we only want to rename the interaction as `Navigation` when the interaction was named using the CSS selector (third priority). Otherwise, if this is removed, the interaction will be named as `Navigation` even if the developer used the attribute `data-ocweb-id`.

draffensperger: Hmm, OK. What if we just gave all interaction spans some temporary name (maybe just the empty string), and then do the logic for what to name the span at the end?

crdgonzalezca: mmm yeah, in that case it won't work as expected. What we can do is to check `rootSpan.name` is not empty, then we will know we are not overwriting the value in `data-ocweb-id`.

packages/opencensus-web-instrumentation-zone/test/test-interaction-tracker.ts
```diff
@@ -260,49 +260,85 @@ describe('InteractionTracker', () => {
...
+        expect(rootSpan.ended).toBeTruthy();
+        expect(rootSpan.duration).toBeGreaterThanOrEqual(SET_TIMEOUT_TIME);
+        expect(rootSpan.duration).toBeLessThanOrEqual(
+          SET_TIMEOUT_TIME + TIME_BUFFER
+        );
+        done();
+      });
+    });
+
+    it('should handle route transition interaction and not rename the interaction as Navigation', done => {
```
draffensperger: Maybe say `'should not rename span as Navigation if it used data-ocweb-id'`?

crdgonzalezca:
In order to be able to know when the name of interaction can change as Navigation, moved the History API patch to a separate module and now the Interaction names are marked whether their name can change or not.

packages/opencensus-web-instrumentation-zone/src/history-api-patch.ts
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function maybeUpdateInteractionName(previousLocationPathname: string) {
+  const rootSpan = tracing.tracer.currentRootSpan;
+  // If for this interaction, the developer did not give any explicit
+  // attibute (`data-ocweb-id`) and the generated name can be replaced,
+  // that means the name might change to `Navigation <pathname>` as this is a
+  // more understadable name for the interaction in case the location
+  // pathname actually changed.
+  const traceId = getTraceId(Zone.current);
```
draffensperger: Rather than using a `Set` to track the interactions, could you just have another property on the `Zone` and an accessor function like `getRenameOnNavigation` or similar?

My thinking is that may be a little cleaner to think about and avoids the need to add/remove elements to the `Set`

crdgonzalezca:
Change the logic to add a flag on the Zone to tell the History API path whether the name can change.

PR #144, opened 2019-07-22
Allow sampling for user interaction traces

crdgonzalezca:
The sampling decision is done from the beginning, that way we decide whether all the interaction traces are sampled or not. Additionally, the sampling decision will be the same for the Initial Load Page, and for doing that, @opencencus/web-initial-load package is created to have all the logic related to the initial load spans and storing the Initial Load Span Context to be able to pass it around and know the sampling decision done for the initial load.


examples/user_interaction/client/src/index.js
```diff
@@ -48,6 +48,8 @@ window.ocAgent = 'http://localhost:55678';
 // For the purpose of this example, send trace header to all hosts.
 window.ocTraceHeaderHostRegex = /.*/;
 window.ocSampleRate = 1.0; // Sample at 100% for test only. Default is 1/10000.
+// Sample user interactions at 100% for test only. Default is 1/10000.
+window.ocUserInteractionSampleRate = 1.0;
```
draffensperger: What would you think about just consolidating this with the `ocSampleRate` parameter such that a single sampling decision is made for whether to trace the initial load and all on-page interactions?

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -60,15 +60,18 @@ export declare interface WindowWithOcwGlobals extends Window {
    * Example 'https://my-oc-agent-deployment.com:55678'
    */
   ocAgent?: string;
+
   /**
-   * If the `traceparent` global variable described above is not present on the
-   * `window`, then a trace sampling decision will be made randomly with the
-   * specified sample rate. If not specified, a default sampling rate is used.
+   * A trace sampling decision to start the interaction tracker and sample all
+   * the interactions or non of them will be made randomly with the specified
```
draffensperger: Nit: "none of them"

packages/opencensus-web-instrumentation-zone/src/interaction-sampling.ts
```diff
@@ -0,0 +1,40 @@
...
+ * decision is done using the `window.ocUserInteractionSampleRate` value or a
+ * default a value of 1/10000.
+ */
+export function interactionTrackerIsSampled(): boolean {
+  const sampleRate =
+    windowWithOcwGlobals.ocUserInteractionSampleRate || DEFAULT_SAMPLE_RATE;
+  // `Math.random` returns a number in the 0-1 range (inclusive of 0 but not 1).
+  // That means we should use the strict `<` operator to compare it to the
+  // sample rate.
+  return !!(Math.random() < sampleRate ? 1 : 0);
```
johnbryan: can this line be simplified to "return Math.random() < sampleRate;" ?

examples/user_interaction/client/src/index.js
draffensperger: Can this line be removed now?

packages/opencensus-web-all/src/export-initial-load.ts
```diff
@@ -66,7 +66,7 @@ export function exportRootSpanAfterLoadEvent() {
 function exportInitialLoadSpans() {
   setTimeout(() => {
     const spanContext = getInitialLoadSpanContext();
-    if (!isSampled(spanContext)) return; // Don't export if not sampled.
+    if (!isSampled()) return; // Don't export if not sampled.
```
draffensperger: Why did you remove the `spanContext` argument here?

packages/opencensus-web-all/src/initial-load-context.ts
```diff
@@ -14,19 +14,19 @@
...
 
 import { WindowWithOcwGlobals } from './types';
 
 const windowWithOcwGlobals = window as WindowWithOcwGlobals;
 
-/**
- * The default trace sampling rate if no `traceparent` and no `ocSampleRate`
- * are specified on the `window`.
- */
-const DEFAULT_SAMPLE_RATE = 0.0001;
```
draffensperger: Why is this default sample rate being removed? Would it be OK to keep it?

packages/opencensus-web-all/src/initial-load-context.ts
```diff
@@ -42,17 +42,16 @@ export function getInitialLoadSpanContext(): SpanContext {
     console.log(`Invalid traceparent: ${windowWithOcwGlobals.traceparent}`);
     return randomSampledSpanContext();
   }
+  // Update the sampling decision with the value coming from the traceparent.
+  const traceparentSamplingDecision = spanContext.options || 0;
```
draffensperger: What role does the `|| 0` play? What would happen if it were removed?

packages/opencensus-web-all/src/initial-load-context.ts
```diff
@@ -42,17 +42,16 @@ export function getInitialLoadSpanContext(): SpanContext {
...
+  updateSamplingDecision(traceparentSamplingDecision);
```
draffensperger: Would it work to just cache the whole initial load span context rather than just the sampling decision?

My thinking is that we will need the trace ID and span ID anyway for adding the links to the initial load span, so we may as well cache the whole object.

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -60,15 +60,11 @@ export declare interface WindowWithOcwGlobals extends Window {
...
   ocAgent?: string;
+
   /**
-   * If the `traceparent` global variable described above is not present on the
-   * `window`, then a trace sampling decision will be made randomly with the
-   * specified sample rate. If not specified, a default sampling rate is used.
+   * RegExp to control what origins will the `trace context header` be sent.
+   * That way the header is not added to all xhrs.
    */
-  ocSampleRate?: number;
```
draffensperger: Why is `ocSampleRate` being removed here?

crdgonzalezca:
The logic has changed following your idea of storing the Span Context. The sampling decision will be the same for the Initial Load Page, and for doing that, the Initial Load Span Context is stored in @opencencus/web-core package to be able to pass it around and know the sampling decision done for the initial load.

packages/opencensus-web-all/src/export-initial-load.ts
```diff
@@ -66,7 +70,9 @@ export function exportRootSpanAfterLoadEvent() {
 function exportInitialLoadSpans() {
   setTimeout(() => {
     const spanContext = getInitialLoadSpanContext();
-    if (!isSampled(spanContext)) return; // Don't export if not sampled.
+    // Set the span context to allow passing it around OpenCensus Web.
+    setInitialLoadSpanContext(spanContext);
```
draffensperger: What would you think about moving this `setInitialLoadContext` over to the `getInitialLoadSpanContext` function? So that it just sets it the first time you get it?

packages/opencensus-web-all/src/export-initial-load.ts
```diff
@@ -66,7 +70,9 @@ export function exportRootSpanAfterLoadEvent() {
...
+    if (!isSampled()) return; // Don't export if initial load is not sampled.
```
draffensperger: Would it make sense to pass the `spanContext` argument here so that it's clear to the reader that the sampling decision is coming from the initial load span context?

packages/opencensus-web-all/src/initial-load-context.ts
```diff
@@ -39,14 +40,22 @@ const DEFAULT_SAMPLE_RATE = 0.0001;
...
+    );
+    if (!spanContextFromTraceparent) {
+      console.log(`Invalid traceparent: ${windowWithOcwGlobals.traceparent}`);
+      spanContext = randomSampledSpanContext();
+    } else {
+      spanContext = spanContextFromTraceparent;
+    }
   }
+  // Set the span context to allow passing it around OpenCensus Web.
+  setInitialLoadSpanContext(spanContext);
```
draffensperger: What would you think about moving `setInitialLoadSpanContext` into this `initial-load-context.ts` file and making it non-exported? My thinking is that the user of the utility should just need to call `getInitialLoadSpanContext` and not need to worry about how it is generated / stored internally.

packages/opencensus-web-core/test/test-initial-load-util.ts
```diff
@@ -18,21 +18,22 @@ import {
   isSampled,
   makeRandomSamplingDecision,
   setInitialLoadSpanContext,
+  initialLoadSpanContext,
```
draffensperger: Rather than exporting the variable, could we export `getInitialLoadSpanContext`?

packages/opencensus-web-all/src/initial-load-context.ts
crdgonzalezca: I think having this adds more complexity and makes less understandable the code, I think that having the setter there is better. Additionally, if the `initialLoadSpanContext` will have a getter, the logic of setting this value from the `initial-load-context` is not good as we have to first get the value and then modify it.

crdgonzalezca: Final decision: Create @opencensus/web-initial-load package to store this logic rather than using the core package.

crdgonzalezca:
Made a bunch of changes. In general the creation of @opencensus/web-initial-load to wrap the logic related to initial load tracing and sharing the span context.
Ready for review.

packages/opencensus-web-initial-load/karma.conf.js
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-const webpackConfig = require('./webpack/dev-package.config');
+const webpackConfig = require('./webpack.config');
```
draffensperger: Would you be willing to keep a basic `karma.conf.js` file in `opencensus-web-all` in case we add tests there in the future?

packages/opencensus-web-initial-load/package.json
```diff
@@ -0,0 +1,71 @@
+{
+  "name": "@opencensus/web-initial-load",
+  "version": "0.0.3",
+  "description": "OpenCensus Initial Load trace",
```
draffensperger: Maybe "OpenCensus Web trace instrumentation for the initial page load"?

packages/opencensus-web-initial-load/src/types.ts
```diff
@@ -0,0 +1,37 @@
...
+export declare interface WindowWithInitialLoadGlobals
+  extends WindowWithLongTasks,
+    WindowWithOcwGlobals {
+  /**
+   * List to collect long task timings as they are observed. This is on the
+   * window so that the code to instrument the long tasks and the code that
+   * exports it can be in different JS bundles. This enables deferring loading
+   * the export code until it is needed.
+   */
+  ocLt?: PerformanceLongTaskTiming[];
```
draffensperger: Is this needed here? I would have guessed that `WindowWithLongTasks` would have this in its definition?

packages/opencensus-web-instrumentation-zone/src/zone-types.ts
```diff
@@ -54,20 +54,8 @@ export type XhrWithOcWebData = HTMLElement &
...
-   * If the `traceparent` global variable described above is not present on the
-   * `window`, then a trace sampling decision will be made randomly with the
-   * specified sample rate. If not specified, a default sampling rate is used.
-   */
-  ocSampleRate?: number;
-
+export declare interface WindowWithInteractionGlobals
+  extends WindowWithOcwGlobals {
   /**
    * RegExp to control what origins will the `trace context header` be sent.
```
draffensperger: Given that the window is a single namespace, what would you think about collapsing all the different `WindowWithXXXGlobals` into a single `WindowWithOcwGlobals` type in the `web-core` package?

packages/opencensus-web-initial-load/karma.conf.js
crdgonzalezca: OK. Added.

packages/opencensus-web-all/src/export-initial-load.ts
```diff
@@ -14,71 +14,4 @@
...
-      spanContext.spanId,
-      spanContext.traceId
-    );
-
-    clearPerfEntries();
-    // Notify that the span has ended to trigger export.
-    tracing.tracer.onEndSpan(root);
-  }, WAIT_TIME_AFTER_LOAD_MS);
-}
+export { exportRootSpanAfterLoadEvent } from '@opencensus/web-initial-load';
```
draffensperger: Could you move this to the `index.ts` file for `opencensus-web-all`?

packages/opencensus-web-initial-load/src/index.ts
```diff
@@ -14,13 +14,6 @@
...
-
-const IS_SAMPLED_BIT = 0x1;
-
-/** Returns whether sampling hint bit of the span context `options` is set. */
-export function isSampled(spanContext: SpanContext) {
-  const options = spanContext.options;
-  if (!options) return false;
-  return !!(options & IS_SAMPLED_BIT);
-}
+export { exportRootSpanAfterLoadEvent } from './export-initial-load';
```
draffensperger: Same here on moving these to `index.ts` if we are just re-exporting them to keep compatibility

crdgonzalezca: I'm not sure to understand this. These exports are already in the `index.ts`

PR #145, opened 2019-07-26
Relate user interaction traces back to the initial page load trace

crdgonzalezca:
- Add `initia_load_trace_id` to every the user interaction root spans and the initial load root span. This allows the customer to search all traces related to a whole navigation through the initial page.
- Add a link to the initial load trace so the customers can easily go back to the initial trace.

Here is an example on how the user interaction trace looks like:
image

### DIFF
--- a/packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
+++ b/packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
@@ -20,6 +20,7 @@ import {
   ATTRIBUTE_HTTP_USER_AGENT,
   ATTRIBUTE_LONG_TASK_ATTRIBUTION,
   ATTRIBUTE_NAV_TYPE,
+  ATTRIBUTE_INITIAL_LOAD_TRACE_ID,
   parseUrl,
   randomSpanId,
   randomTraceId,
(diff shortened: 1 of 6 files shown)


packages/opencensus-web-instrumentation-perf/src/initial-load-root-span.ts
```diff
@@ -92,6 +93,7 @@ export function getInitialLoadRootSpan(
   root.annotations = getNavigationAnnotations(perfEntries);
   root.attributes[ATTRIBUTE_HTTP_URL] = navigationUrl;
   root.attributes[ATTRIBUTE_HTTP_USER_AGENT] = navigator.userAgent;
+  root.attributes[ATTRIBUTE_INITIAL_LOAD_TRACE_ID] = root.traceId;
```
draffensperger: Could you add that this is included to enable trace search by attribute to find an initial load trace and its interaction traces via a single attribute query?

packages/opencensus-web-instrumentation-zone/src/on-page-interaction-stop-watch.ts
```diff
@@ -56,6 +60,16 @@ export class OnPageInteractionStopwatch {
...
     rootSpan.addAttribute(ATTRIBUTE_HTTP_USER_AGENT, navigator.userAgent);
+    const initialLoadSpanContext = getInitialLoadSpanContext();
+    rootSpan.addAttribute(
+      ATTRIBUTE_INITIAL_LOAD_TRACE_ID,
+      initialLoadSpanContext.traceId
+    );
+    rootSpan.addLink(
+      initialLoadSpanContext.traceId,
+      initialLoadSpanContext.spanId,
+      LinkType.CHILD_LINKED_SPAN
```
draffensperger: I think I got the type backwards, I think it should be `PARENT_LINKED_SPAN` since the linked span is the current span's parent. 

Relevant doc: https://github.com/census-instrumentation/opencensus-proto/blob/d4d9f6326c6526ba036327dd0237a468de592f04/src/opencensus/proto/trace/v1/trace.proto#L266

PR #147, opened 2019-07-30
Add @opencensus/web-instrumentation-zone-peer-dep

crdgonzalezca:
Add this new package allow customers to choose a package with `Zone.js` as hard dependy or as peer dependency. This allows to applications in frameworks like React to avoid adding that dependency, also, for apps already depending on `Zone.js` this is not added, that is the case for Angular apps.


packages/opencensus-web-instrumentation-zone-peer-dep/README.md
```diff
@@ -0,0 +1,60 @@
...
+[![Gitter chat][gitter-image]][gitter-url]
+
+*For overview and usage info see the main [OpenCensus Web readme][oc-web-readme-url].*
+
+This package generates OpenCensus Web model spans based on user interactions 
+coming after initial page loads. This uses the library 
+[Zone.js](https://github.com/angular/zone.js/) to generate those spans.
+
+This package adds `Zone.js` as a peer dependency as the application 
+might already be dependet on it. This is the case, for example, with
```
draffensperger: Nit: depdendent

crdgonzalezca:
This PR does not update the `Zone.js` library to the last version and the problem is related to that version. I am planning on doing this for a different PR as I consider this is big enough.

PR #148, opened 2019-07-31
Refactor packages and rename @opencensus/web-all to @opencensus/web-scripts

crdgonzalezca:
This PR has several changes in packages to change the way OC Web is used:
* Rename @opencensus/web-all to @opencensus/web-scripts so this package now is in charge of bundling OC Web to allow importing it as <script> tag.
* New entry points are created: `tracing-all-with-zone` and `tracing-all-without-zone`, this allows the customers to select the desired script depending on whether their app already has `Zone.js` or not.
* Update a little the documentation. I am planning to update the whole documentation in another PR as I consider this is big enough.


packages/opencensus-web-scripts/README.md
```diff
@@ -45,6 +46,28 @@ these two bundles to do instrumentation first and then an export more lazily:
     downloaded and parsed doesn't need to compete with resources for the main
     page load.
 
+If your application already depends on `Zone.js`, the case for `Angular`, use:
+
+* **`./dist/tracing-all-without-zone.js`** - This script combines all the functionality of
```
draffensperger: What about calling this `tracing-all-zone-peer-dep`? The phrase "without-zone" makes me think it might not be dependent on Zone.js at all, but in fact it is, just indirectly

packages/opencensus-web-scripts/README.md
```diff
@@ -45,6 +46,28 @@ these two bundles to do instrumentation first and then an export more lazily:
...
+    OC Web, including the `initial-page-load` and `user interaction tracker`. That
```
draffensperger: Are the strings "initial-page-load" and "user interaction tracker" meant to refer to code symbols? (That's what the backticks usually mean). If the are just being used for emphasis, consider using * around them to make them italic and removing the dashes in the initial-page-load one.

crdgonzalezca:
Updated the CHANGELOG. Ready for review.

PR #151, opened 2019-08-05
Fix generation of `tracing-all-with-zone` bundle script

crdgonzalezca:
Fix `tracing-all-with-zone.js` bundle script to take the `Zone.js` code with it.


packages/opencensus-web-instrumentation-zone/package.json
```diff
@@ -64,5 +64,5 @@
     "@opencensus/web-instrumentation-zone-peer-dep": "^0.0.3",
     "zone.js": "~0.9.1"
   },
-  "sideEffects": false
+  "sideEffects": true
```
draffensperger: Nice catch here! And a good reason for this to be a separate package from the `-peer-dep` one!

PR #162, opened 2019-08-12
Fix link to bundle with zone-peer-dep

backjo:
This change is [Reviewable](https://reviewable.io/reviews/census-instrumentation/opencensus-web/162)


backjo:
@googlebot I signed it!

PR #173, opened 2019-08-29
OpenCensus Web B3 Propagation

edvinasbartkus:
Creates a new web package called `opencensus-web-b3-propagation`.
The way web b3 propagation works is identical to nodejs/b3-propagation. In this package, it includes `web-core` for internal references (for example for random trace id generation).

`startTracing` is changed to accept `options = { propagation: Propagation }`.

By default `opencensus/web-propagation-tracecontext.TraceContextFormat` will be used but it can be replaced with `opencensus/web-propagation-b3.B3Format`

@draffensperger am I on the right path? :)

### DIFF
--- a/packages/opencensus-web-instrumentation-zone-peer-dep/src/interaction-tracker.ts
+++ b/packages/opencensus-web-instrumentation-zone-peer-dep/src/interaction-tracker.ts
@@ -19,6 +19,7 @@ import {
   tracing,
   SpanKind,
   RootSpan,
+  Propagation,
 } from '@opencensus/web-core';
 import { AsyncTask, InteractionName } from './zone-types';
 import {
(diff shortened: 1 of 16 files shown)


edvinasbartkus:
@googlebot I signed it!

packages/opencensus-web-instrumentation-zone-peer-dep/src/interaction-tracker.ts
```diff
@@ -35,7 +36,13 @@ declare const Zone: ZoneType & { prototype: Zone };
 // Delay of 50 ms to reset currentEventTracingZone.
 export const RESET_TRACING_ZONE_DELAY = 50;
 
+export interface TrackerOptions {
+  propagation?: Propagation;
+}
+
 export class InteractionTracker {
+  //
```
draffensperger: Could you either remove this or fill out the comment?

edvinasbartkus: Removed.

PR #279, opened 2020-05-21
Add package.json metadata for ES6 module imports

ad-m:
In Azure/azure-sdk-for-js#9060 bterlson wrote: 

> Additionally, @opencensus/web-types doesn't have the correct package.json metadata to contain ES6 module imports.

I have no experience in ES6 modules (it's not available in any LTS version of NodeJS), so please kindly careful reviews.

### DIFF
--- a/packages/opencensus-web-types/package.json
+++ b/packages/opencensus-web-types/package.json
@@ -30,6 +30,7 @@
   "engines": {
     "node": ">=6.0"
   },
+  "type": "module",
   "files": [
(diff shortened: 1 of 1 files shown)


ad-m:
@googlebot I signed it!

bterlson:
@ad-m I think a better fix is to update the tsconfig.json module entry to "commonjs" so the package doesn't need ES modules.

ad-m:
@bterlson, review comment applied.

